PR #3, opened 2020-06-29
Fixed no auth handshake, allowing users to connect if vnc auth disabled

zach-r-long:
Allows for a no auth connections when specified in configuration file.


mdevaev:
Sup. Could you explain the path to reproducing of this bug? I thought it was working.

zach-r-long:
Seems issue actually lies with encoding. Is raw not supported anymore? Work previously in the May timeframe. Will close PR.

mdevaev:
RAW did never work...

PR #4, opened 2020-08-15
Update navbar-msd.pug

tryallthethings:
Changed ambiguous wording for image upload buttons.


mdevaev:
Thank you, it will be easier to understand. But there are two problems:

1) the Second label is too wide, it is better to just "Upload".

2) Changing pug is not enough, the interface is completely static and pug files are just sources for html that needs to be regenerated using `make pug`. Well, just consider for future patches.

I'll fix it myself.

tryallthethings:
1. I tested the labels in Chrome before committing them and they looked good on my end:
image
Anyway - just upload is fine as well :smiley: 
2. Right. My understanding was that the html was generated at runtime. Will keep that in mind! Thanks for fixing!

PR #6, opened 2020-09-01
Make KVMD compatible with Python 3.7

OlegGirko:
Using new features provided by Python 3.8 is nice, but there are popular Linux distributions (like Debian / Raspbian) that still use Python 3.7.

Only two 3.8-specific features are used in KVMD:
* assignment expressions using `:=` operator,
* `{variable=}` substitution in f-strings.

This set of changes makes KVMD compatible with Python 3.7 by rewriting its code to not use these features. I tried to do it carefully, without sacrificing readability.

I've tested these changes thoroughly. I have KVMD with these changes working flawlessly on Raspberry Pi 4 with Raspbian.

### DIFF
--- a/kvmd/aiotools.py
+++ b/kvmd/aiotools.py
@@ -177,5 +177,6 @@ async def run_region_task(
 
     if entered.done():
         return
-    if (exc := task.exception()) is not None:
+    exc = task.exception()
+    if exc is not None:
         raise exc
--- a/kvmd/apps/kvmd/api/auth.py
+++ b/kvmd/apps/kvmd/api/auth.py
@@ -45,7 +45,8 @@ _COOKIE_AUTH_TOKEN = "auth_token"
 
 async def check_request_auth(auth_manager: AuthManager, exposed: HttpExposed, request: Request) -> None:
     if exposed.auth_required and auth_manager.is_auth_enabled():
-        if (user := request.headers.get("X-KVMD-User", "")):
+        user = request.headers.get("X-KVMD-User", "")
+        if user:
             user = valid_user(user)
             passwd = request.headers.get("X-KVMD-Passwd", "")
             set_request_auth_info(request, f"{user} (xhdr)")
@@ -53,7 +54,8 @@ async def check_request_auth(auth_manager: AuthManager, exposed: HttpExposed, re
                 raise ForbiddenError()
             return
 
-        elif (token := request.cookies.get(_COOKIE_AUTH_TOKEN, "")):
+        token = request.cookies.get(_COOKIE_AUTH_TOKEN, "")
+        if token:
             user = auth_manager.check(valid_auth_token(token))
             if not user:
                 set_request_auth_info(request, "- (token)")
@@ -61,7 +63,8 @@ async def check_request_auth(auth_manager: AuthManager, exposed: HttpExposed, re
             set_request_auth_info(request, f"{user} (token)")
             return
 
-        elif (basic_auth := request.headers.get("Authorization", "")):
+        basic_auth = request.headers.get("Authorization", "")
+        if basic_auth:
             if basic_auth[:6].lower() == "basic ":
                 try:
                     (user, passwd) = base64.b64decode(basic_auth[6:]).decode("utf-8").split(":")
--- a/kvmd/apps/kvmd/api/streamer.py
+++ b/kvmd/apps/kvmd/api/streamer.py
@@ -56,11 +56,12 @@ class StreamerApi:
 
     @exposed_http("GET", "/streamer/snapshot")
     async def __take_snapshot_handler(self, request: Request) -> Response:
-        if (snapshot := await self.__streamer.take_snapshot(
+        snapshot = await self.__streamer.take_snapshot(
             save=valid_bool(request.query.get("save", "false")),
             load=valid_bool(request.query.get("load", "false")),
             allow_offline=valid_bool(request.query.get("allow_offline", "false")),
-        )):
+        )
+        if snapshot:
             if valid_bool(request.query.get("preview", "false")):
                 data = await self.__make_preview(
                     snapshot=snapshot,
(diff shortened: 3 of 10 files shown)


mdevaev:
There are several reasons why I would not want to accept this patch.
1. In addition to the syntactic sugar, there are a number of changes between 3.7 and 3.8 to asyncio, which is where the entire KVMD is built. Accepting the patch into the official branch means that I will have to commit to keeping KVMD running on 3.7. I can't just say "yes, it works on 3.7" and not test it. Since I'm currently only testing on 3.8 and using some hack for asyncio, this might break down. The bad news is that I don't remember exactly what broke down in 3.7.
2. 3.8 will appear in repositories at some point. Isn't there some deadsnakes for this?

mdevaev:
Sorry

OlegGirko:
I understand that you want to support only platform that you use. But Raspbian is much more widely used distro, so you do a big disservice to adoption of your software by ignoring it.

I don't know what `asyncio` problem you mean, but `kvmd` works flawlessly on Raspbian with my patches. And, as you see, my patches are minimal, just removing syntactic sugar allowed by Python 3.8 that is not so needed for readability.

The only commitment you'll have after accepting my patches is to not use assignment expressions in your further changes. I already do testing by building Debian packages running `kvmd` that controls my desktop computer (that is also my main server). If your next change breaks compatibility with Raspbian, I'll definitely send you a pull request.

If you really want to somehow indemnify yourself from moral responsibility, you can leave a note in `README` that you test `kvmd` only in Arch Linux, and all others who use it on other distros do it on their own risk. You can also mention me as a maintainer of Debian (and Fedora) packages.

And no, there are no indications that Python 3.8 is going to appear in Debian 10 stable. We have to wait a few years until Debian 11 (and then Raspbian based on Debian 11) is released.

Python 3.8 is already in Debian unstable, but It's not installable on Debian 10 because it was built with newer versions of `glibc` and `libgcc` and requires them. And there is strong recommendation on Debian site against creating FrankenDebian by mixing packages from stable and unstable.

mdevaev:
Okay, your argument sounds reasonable. I need to think a little.

mdevaev:
By the way, I wonder how you managed to configure pikvm for raspbian, because in addition to kvmd and ustreamer, you need a lot of other dependencies, as well as configs. Are there any published sources for your packages and/or instructions?

OlegGirko:
> By the way, I wonder how you managed to configure pikvm for raspbian, because in addition to kvmd and ustreamer, you need a lot of other dependencies,

All necessary dependencies are already available in Debian / Raspbian repos.

> as well as configs.

For now I've packaged `kvmd` the way that configs have to be installed manually. All default configs for all supported variants are installed in `/usr/share/kvmd/configs.default` directory and have to be placed manually into `/etc/kvmd` directory.

> Are there any published sources for your packages and/or instructions?

I use OBS to build packages for Fedora 32 aarch64 and Debian 10 armv7l: https://obs.infoserver.lv/project/monitor/pikvm
You can find sources there: https://obs.infoserver.lv/project/show/pikvm

To use package repo that is generated by OBS, create `/etc/apt/sources.list.d/pikvm.list` file with the following contents:
```
deb https://obs.infoserver.lv/repos/pikvm/Debian_10/ /
```

For Fedora, create `/etc/yum.repos.d/pikvm.repo` file with the following contents:
```
[pikvm]
name=Open and cheap DIY IP-KVM based on Raspberry Pi (Fedora_$releasever)
type=rpm-md
baseurl=https://obs.infoserver.lv/repos/pikvm/Fedora_$releasever/
gpgcheck=1
gpgkey=https://obs.infoserver.lv/repos/pikvm/Fedora_$releasever/repodata/repomd.xml.key
enabled=1
```

Although I've packaged `kvmd` for Fedora, I didn't test it yet. I need another Raspberry Pi for that, and I'll get it in a few weeks.

mdevaev:
That's cool. I will accept these changes later today when I finally Wake up. I want to refactor some things, but you don't have to worry about it.

By the way, is ustreamer going to support OMX and wiringpi? At first glance, no. OMX is necessary for some configurations, and the second one is just a useful feature. I suggest add two dependencies `libraspberrypi-dev` and `wiringpi` and build as follows: `make WITH_OMX=1 WITH_GPIO=1`.

OlegGirko:
> By the way, is ustreamer going to support OMX and wiringpi?

OMX requires proprietary Broadcom libraries that are Raspberry Pi specific and not available in Debian repos.
WiringPi is [deprecated and not developed anymore](http://wiringpi.com/wiringpi-deprecated/) and is also Raspberry Pi specific.

I'm building using regular Debian repos because these packages can eventually be useful for other ARM-based platforms and distros (like Armbian). It would be great if `kvmd` and `ustreamer` were changed to use more portable (and modern) interface to GPIO, like `libgpiod` / `python3-libgpiod`. May be I'll take a look at that later.

mdevaev:
https://github.com/pikvm/pikvm/issues/65

mdevaev:
> OMX requires proprietary Broadcom libraries that are Raspberry Pi specific and not available in Debian repos.

Okay, but JFYI: the operation of the HDMI-CSI bridge will be extremely problematic without GPU compression due performance.

OlegGirko:
> > OMX requires proprietary Broadcom libraries that are Raspberry Pi specific and not available in Debian repos.
> 
> Okay, but JFYI: the operation of the HDMI-CSI bridge will be extremely problematic without GPU compression due performance.

Then this part can be made as an optional plugin, for example.

mdevaev:
It is problematic to make OMX as a plugin, since its presence changes the behavior of some other components. I would rather have a nonfree version with additional dependencies in the repositories.

aggieNick02:
Just a comment and thanks to both of you. I'm also playing with running this in Raspberry PI OS. I had no idea I would have run into a few problems a few weeks ago, and just realized seeing this issue that I was very fortunate in the timeframe I learned about and started playing with Pi-KVM. Thanks!

PR #10, opened 2020-09-20
Added missing key on non us keyboards

mfunkey:
Some non us keyboards use an extra key for less, greater and bar (German).

I've added a blank key to the virtual keyboard since i don't know if this key represents anything on a us keyboard.

Tested with otg, hid-usb and hid-ps2

### DIFF
--- a/hid/src/ps2/keymap.h
+++ b/hid/src/ps2/keymap.h
@@ -144,5 +144,6 @@ INLINE void keymapPs2(uint8_t code, Ps2KeyType *ps2_type, uint8_t *ps2_code) {
 		case 103: *ps2_type = PS2_KEY_TYPE_REG; *ps2_code = 112; return; // KEYPAD_0
 		case 104: *ps2_type = PS2_KEY_TYPE_REG; *ps2_code = 113; return; // KEYPAD_DOT
 		case 105: *ps2_type = PS2_KEY_TYPE_SPEC; *ps2_code = 94; return; // KEY_POWER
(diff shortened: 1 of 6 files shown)


mdevaev:
This patch conflicts with other AT1 keys (see autotests). We must resolve it. Could you explain about NON_US key in details? What is it? How it looks in your keyboard?

mfunkey:
In germany we use 105-key keyboard instead of 104-key.

The missing key is the one between left-shift and the y-key(bottom left)
https://en.wikipedia.org/wiki/German_keyboard_layout#/media/File:KB_Germany.svg

For more information see scan code "0x56":
https://www.win.tue.nl/~aeb/linux/kbd/scancodes-1.html#ss1.4

I've switch my machines keyboard layout to us and this key also types the less and greater key.

I'm pretty new to your project and python.
Do you have any idea how to work around this conflicting AT1 keys?
What are they used for?
Do these AT1 keys have to be unique or ist this just a "dict" limitation?

mdevaev:
These must be unique mappings. Okay, I understand. I'll read about layouts and find out how it should be done.

mfunkey:
Okay, i'll look into it as well.

mdevaev:
Okay, I accept that, but I'll redo the key layout and something else

mdevaev:
I placed this key in the corner of the virtual keyboard to avoid making the US layout a Frankenstein's monster. The full Non-US layout must contain other changes. Let's leave this for future support of the alternative layout.

PR #15, opened 2020-10-02
added ssl in test mode

arykov:
Adds ssl, although does not force it in test mode(when running using make run). So you can access using SSL on 8443 or no SSL on 8080
Rationale: certain browser functionality(for example keyboard lock api) is only available over SSL.


arykov:
Jumped the gun. Broke http

arykov:
Http should be working now

mdevaev:
The new kernel release with critical fixes was released today, so now I will make bugfixes and build a new version. This change will be included in the next.

mdevaev:
Okay, I tried to implement it differently. To generate certificates, I used the kvmd-gencert script (extra testing) and did without editing nginx.conf in the repository. It seems to be working well now. Thank you for coming up with these ideas.

mdevaev:
UPD: https://github.com/pikvm/kvmd/commit/923320da5f87d6081824641468b3777d7f1e811f

PR #21, opened 2020-12-15
Programmatically reconnect otg device on hid reset

Yura80:
This solves the problem with some KVMs, like SeaWit and potentially enables absolute/relative mouse mode switching on the fly for OTG.


mdevaev:
Implemented as pseudo-gpio plugin: https://github.com/pikvm/kvmd/commit/570dd04e29f5c597cf1fdcf9eca8f9f9877d243d

PR #25, opened 2021-01-26
Fix spelling error in web interface

owendaprile:
Got one of the smallest pull requests ever here.
Just noticed this while messing around with the web interface.


mdevaev:
Thanks!

PR #27, opened 2021-02-07
More web interface spelling fixes + a little improvement

owendaprile:
Came across a few more spelling errors in the web interface and fixed them.

Also updated the paste dialog to change depending on whether there's 1 or 0,2+ characters.

Also disabled the warning dialog when leaving the page if an unexpected logout has occurred. It was a little annoying to me when I was messing with my build and kept restarting the server which showed the prompt but still had the browser warning.


owendaprile:
image
Here's the test results on my computer because of the image problems.

mdevaev:
The Arch guys broke glibc. I'll accept it. Thanks!

PR #37, opened 2021-03-28
zh_CN translate

wnark:
Tentative Chinese translation


mdevaev:
Wait, did you localize HTML files, not PUG?

wnark:
Yes, I just translated the html and just need overwrite the web folder. (I'm lazy and don't want to overwrite files one by one)

mdevaev:
You had to copy and edit the PUG files from which the HTML is generated.
For example: https://github.com/pikvm/kvmd/blob/master/web/index.pug

mdevaev:
I'm sorry, but I can't apply it in this form.

Issue #14, opened 2020-10-01
Feature - maximizing window


arykov:
As discussed [here](https://discord.com/channels/580094191938437144/580094193834131456/758069937112744136) I am looking for a way to maximize use of screen real estate by maximizing streaming window.

arykov:
Here's my prototype of scaling and full screen code: https://github.com/arykov/kvmd/commits/master
Could you have a look to see if you find anything offensive?
Known issues:
1) and most important I am still waiting for parts to test this on real build. So there are a lot of unknowns for mouse alignment and keyboard
2) Testing of video resolution change has not been done.  "Auto resize stream" off does not work(can we get rid of it so it is always on?). The best way of testing it would be using v4l2 loopback, which I have not been able to make work yet(https://discord.com/channels/580094191938437144/580094193834131456/760483352246616105). 
3) ESC in full chrome screen mode cannot be intercepted and will
4) Only played with it in Chrome on W10. There clearly will be issues related to IPad resolutions since there are non overridable values, that will prevent hiding top of the window

mdevaev:
Work is underway in this branch: https://github.com/pikvm/kvmd/tree/wm

arykov:
Take 2

mdevaev:
Done

PR #39, opened 2021-04-08
Window maximization feature

arykov:
Caught up to implement feature described in issue #14
- Tested on IPad mini IOS 12. Seems to be working
- Edited html directly. Have not touched pugs
- Did not have a chance to test low resolution screens on IPad. Might be too small(probably existing issue if the case). Not a problem with regular browsers, since can simply resize or maximize or whatever. But IPad is special. Need to either add resizing or rely on scaling, but somehow keep keyboard in the original size.


mdevaev:
Continued here: #40

PR #47, opened 2021-05-12
Plugin support for TESMART switch - tesmart.py

davideshay:
Adding support for TESMART switch.
Works over network protocol only via socket connections. Supports setting IP address and port from override.yaml.
Will send parallel documentation patch to pikvm/pikvm.


mdevaev:
The module will require refactoring

davideshay:
Submitted subsequent PR with lint/flake fixes

PR #48, opened 2021-05-14
Fixed lint/flake issues in tesmart.py

davideshay:


mdevaev:
I didn't just mean that. You used blocking sockets, which can cause the process to hang unpredictably on the execution of a network request. It was necessary to use asyncio. Don't worry, I'll rewrite it myself.

PR #49, opened 2021-05-18
fix socket timeout/zero based issues in tesmart

davideshay:
Minor changes in tesmart.py testing on switch:
* Was getting an error that only a 0 timeout is allowed on transport sockets -- also I think making it 0 makes it non-blocking
* Made everything 0 based -- was only accepting ports 1-15 as written. I think this deserves some consideration, as it is confusing to create your override.yaml file saying that what you have in HDMI switch 5 needs to be configured as pin 4.
* When 0 based, you need to add 1 to the write command -- the tesmart "API" is zero based on read and 1 based on write. It was subtracting 1 on write, which is never correct.

### DIFF
--- a/kvmd/plugins/ugpio/tesmart.py
+++ b/kvmd/plugins/ugpio/tesmart.py
@@ -80,12 +80,12 @@ class Plugin(BaseUserGpioDriver):  # pylint: disable=too-many-instance-attribute
         }
 
     def register_input(self, pin: int, debounce: float) -> None:
-        if not (0 < pin < 16):
+        if not (0 <= pin < 16):
             raise RuntimeError(f"Unsupported port number: {pin}")
         _ = debounce
 
     def register_output(self, pin: int, initial: Optional[bool]) -> None:
-        if not (0 < pin < 16):
+        if not (0 <= pin < 16):
             raise RuntimeError(f"Unsupported port number: {pin}")
         _ = initial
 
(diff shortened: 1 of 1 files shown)


mdevaev:
What happens if you comment out the line about timeout at all?

I agree that counting from zero is a dubious solution, but this is how the other modules behave, so it should be consistent.

mdevaev:
Test master again please

davideshay:
Works well, and yes the timeout was not needed ultimately.
My only comment on this point is that as designed it does a lot of polling
of the switch to check the current port. Might be OK, but given the
instability I've seen in the interface I hate to ping it more than
necessary. In my prior iteration, I only checked at startup and after I
changed inputs via the app.

mdevaev:
I did this to avoid possible out-of-sync states when you manually clicked the switch button on kvm and the interface shows the old value. What does this instability look like? I did a polling rate of 5 seconds, isn't that enough?

davideshay:
I thought about that in terms of manually clicking the buttons as a
possible reason to do that. Just leave it as is for now and unless it
crashes the switch over the long haul no need to revisit.

mdevaev:
I don't quite understand what you mean. Leave the code as it is now, or remove polling altogether? Maybe it's better not to remove it, but to increase the interval to 10 seconds?

mdevaev:
Ok, changed to 10

PR #54, opened 2021-08-08
Feature/ctrl in combinations

reallaborious:
Added:
- an ability to catch combinations of keys like CTRL-C in web interface (tested for Chrome/Chromium in Ubuntu 18.04).
- switch to turn off the ability to catch ctrl combinations
- saving of switches' statuses (for all, not only for ctrl combinations) in cookies.

### DIFF
--- a/web/kvm/index.html
+++ b/web/kvm/index.html
@@ -226,6 +226,15 @@
                 </div>
               </td>
             </tr>
+            <tr>
+              <td>Ctrl combination's detect:</td>
+              <td align="right">
+                <div class="switch-box">
+                  <input type="checkbox" id="hid-ctrl-combinations-switch">
+                  <label for="hid-ctrl-combinations-switch"><span class="switch-inner"></span><span class="switch"></span></label>
+                </div>
+              </td>
+            </tr>
           </table>
           <hr>
           <div class="buttons buttons-row">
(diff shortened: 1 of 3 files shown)


mdevaev:
Ctrl+c works now without this, isn't it? Or I'm missing something

reallaborious:
Could you tell please, how long ago was it implemented? I installed os in my pikvm from c9ef3bf7296ecc98c025c6cd49a7ca1750bf6f21 (04/Jule/2021) of https://github.com/pikvm/os, and the only way how I can send ctrl-c to the remote system is with a virtual keyboard, what is really inconvenient when you are working with Linux machines.

mdevaev:
I will check and review it today

Вс, 8 авг. 2021 г. в 09:38, Vladimir K ***@***.***>:

mdevaev:
Just tested Ctrl+C from chromium/linux, everything working :/

reallaborious:
The thing is that all js scripts in my browsers can't catch the pressing of a single ctrl key only, but they catch ctrl with any key, including ctrl-w, ctrl-s etc. And the web interface of kvmd is processing pressing of single key only. I mean that for example it sends to remote server not shif+c, but pressing of Shift, and pressing of c, and after that releasing of both keys.
Anyway, my feature can be turned off with a switch. I suspected that it won't be useful for everyone.

mdevaev:
Please try to find out what is the reason for this behavior, I would prefer not to make some crutches, but to fix the bug.

mdevaev:
It looks like a local problem. No one else is reproducing.

PR #62, opened 2021-09-04
add hue.py ugpio plugin

semool:


mdevaev:
There are a number of serious problems with this code.
1. You are using `requests` (blocking library) while the plugin should use asynchronous `aiohttp`. The current omplementation will cause the entire code to slow down.
2. The `run()` method is not implemented properly. You need to use background polling of the status of ports.

But you don't have to worry about it. Your implementation does the main thing - it shows how to use the API. I will rewrite the plugin myself as needed.

mdevaev:
Do I understand correctly that there can be several devices in one usernme and now you need to create a separate driver instance for each of them?

semool:
Yes, the username is a apikey. Dont know why hue called it username.

mdevaev:
I see. Bad API design is bad.

I'm almost done.

mdevaev:
I made the alternative implementation. Please test it. You need to specify `url` and `token` (i.e. username) params.

mdevaev:
https://github.com/pikvm/kvmd/blob/master/kvmd/plugins/ugpio/hue.py

semool:
Ok also
Username - > token
IP - > URL ; why? the http:// is always the same
and device - > Pin?

semool:
i dont understand the code i think. i have added:
`    gpio:
        drivers:
            hue:
               type: hue
               url: http://192.168.1.7
               token: YG-xxxxxxxxxxxx
        scheme:
            ch7_plug_button:
                driver: hue
                pin: 32
                mode: output
                switch: false
            ch7_plug_led:
                driver: hue
                pin: 32
                mode: input
`

led and button are both disabled.

mdevaev:
Please update not only your modele, but the entire kvmd from git. After that please show `journalctl -u kvmd`

semool:
I have no idea how to do this.

mdevaev:
```
rw
pacman -Syu
git clone --depth=1 https://github.com/pikvm/kvmd
cp -r kvmd/kvmd /usr/lib/python3.9/site-packages/
reboot
```

semool:
OK i have done this. Its not working. After the reboot the plug i have configured is switched to off. The Button is ok but he dont do nothing.

```
Sep 17 15:27:01 pikvm kvmd[7181]: aiohttp.access                    INFO --- [admin (token) / 192.168.1.33] 'POST /gpio/pulse?channel=ch7_plug_button HTTP/1.0' => 200; size=199 --- referer='https://192.168.1.20:4433/kvm/'; user_agent='Mozilla/5.0 (W>
Sep 17 15:27:01 pikvm kvmd[7181]: kvmd.apps.kvmd.ugpio              INFO --- Pulsed Output(ch7_plug_button, driver=Hue(hue), pin=32) with delay=0.10
```

mdevaev:
I probably made a bug somewhere. Is there a Hue protocol emulator somewhere so that I can use it?

semool:
i have played around a bit with your code. The def prepare part switched the plug to offline on kvmd start. I removed this def.
Than i get a plug direct on my side.
Plug led in webif shows correct state -> ON
When i hit the toggle button when the plug is on -> plug goes offline, led is off
When i hit the toggle button again the plug switches in a really short time (1sec?) on and immediatly back to off.

mdevaev:
You don't need to use a button, but a switch. A button is a single click. I have reworked your code so that you can use the switch.
Try this:
```yaml
        drivers:
            hue:
               type: hue
               url: http://192.168.1.7
               token: YG-xxxxxxxxxxxx
        scheme:
            ch7_plug_button:
                driver: hue
                pin: 32
                mode: output
                switch: true
                pulse:
                    delay: 0
            ch7_plug_led:
                driver: hue
                pin: 32
                mode: input
```

semool:
You are right. It was my fault. I should have read the manual pages more accurate.
I have made the changes to switches with the initial option and all is working fine now.
Thank you for your patience and effort.
hue

mdevaev:
No problem! The plugin will be released in the next stable build.

PR #63, opened 2021-09-10
(window.css): correct window border behavior

ly4096x:
prevent window size change between active/inactive.
enable borderless full screen mode.

### DIFF
--- a/web/share/css/kvm/stream.css
+++ b/web/share/css/kvm/stream.css
@@ -53,7 +53,7 @@ div.stream-box-offline::after {
 	background: radial-gradient(transparent 20%, black);
 }
 div.stream-box-mouse-enabled {
-	cursor: url("../../svg/stream-mouse-cursor.svg"), pointer;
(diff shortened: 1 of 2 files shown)


ly4096x:
Looks like removing the border is creating a mouse position glitch by 1px. Need help on this. Thanks.

ly4096x:
Resolved. It's not a problem of changing border, but a problem that has always been there. Centering the local cursor resolves this.

mdevaev:
I would prefer to keep the blue border as sometimes the window can lose focus even in full-screen mode. As for the cursor - this is a good change, let's merge just that.

ly4096x:
I agree that the border can be useful in your case. For me I always use a dual monitor setup so that was not a problem for me. I can try to find a way to keep the border but stop window resizing when getting focus. What to you think?

mdevaev:
Good idea.

mdevaev:
I took the coordinate correction from here and made the blue dot transparent. If you find a way to fix the screen frame, please make another PR.

PR #64, opened 2021-09-14
Add tooltips to SysRq shortcuts

mh166:
Provide a helpful tooltip of what that SysRq shortcut does. As most people don't often (if ever) use SysRq keys, this will be a handy quick reference. Further details can still be gathered from the linked kernel.org article.


mdevaev:
Thanks!

PR #66, opened 2021-09-16
Implemented OCR screenshot.

shantur:
Implement ocr screenshot api. This api takes a screenshot and runs OCR on it and returns all the visible text.

Please merge it on top of GPIO recording #65

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -54,6 +54,7 @@ depends=(
 	python-dbus
 	python-pygments
 	python-pyghmi
+	python-pytesseract
 	python-pam
 	"python-pillow>=8.3.1-1"
 	python-xlib
--- a/kvmd/apps/kvmd/api/streamer.py
+++ b/kvmd/apps/kvmd/api/streamer.py
@@ -31,15 +31,18 @@ from PIL import Image
 from ....validators.basic import valid_bool
 from ....validators.basic import valid_int_f0
 from ....validators.kvm import valid_stream_quality
+from ....validators import check_string_in_list
 
 from .... import aiotools
 
-from ..http import UnavailableError
+from ..http import HttpError, UnavailableError
 from ..http import exposed_http
 from ..http import make_json_response
 
 from ..streamer import StreamerSnapshot
 from ..streamer import Streamer
+
+import pytesseract
 
 
 # =====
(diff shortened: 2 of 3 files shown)


bbeaudoin:
Interesting bit of code, @shantur. It doesn't look like bad code, nice idea for automation but there's a plugin architecture for GPIO handling that I believe could be used instead. Basically an API extension rather than modifying the streamer itself.

To define the driver: https://github.com/pikvm/pikvm/blob/master/pages/gpio.md
plugins would go here but are less well defined: https://github.com/pikvm/kvmd/tree/master/kvmd/plugins

Ideally this would be something one could choose to enable/disable rather than be on all the time.

bbeaudoin:
Combining both PRs to include GPIO recording and OCR Screenshot was, indeed, confusing. That can be handled by a GPIO driver that can be configured externally while screenshots and recordings probably do belong in the screenshot API.

While I might like to see a separate API, there would be unnecessary duplication of code that must be maintained later so extending the API with a related function is appropriate. A feature flag would be nice but, frankly, anyone with access to Pi-KVM or the screenshot API can access the information anyway so my concerns were mostly around the code that fell outside of the scope of the feature request the PR was addressing.

The only thing I take issue with is returning "500 Internal Server Error" rather than "204 No Content" or "510 Not Extended". The generic 500 response should be reserved for errors where the server was unable to perform, such as the API crashing or other fatal error. This can be easily confused with critical errors whereas the other errors are more descriptive. Especially when browser vendors try to produce "helpful" error message screens that obscure the actual server message when such response codes are given.

1. 204 No Content suggests the request and server response was appropriate but there is no content (no identifiable text), per the specifications while there was no content, the request was successful and the headers may be useful.
2. 510 Not Extended suggests it is possible for the server to complete the task if the appropriate language or other configuration can be provided to enable text recognition of the on-screen language. This is most appropriate when sending a textual description of the error but can be confusing if the message was produced in error or the screenshot had no text content.

My suggestion would be to use code 510 so the error and list body can be returned in the body rather than try to include the full message in the headers.

mdevaev:
I made an alternative implementation based on ctypes and with image cropping.
https://github.com/pikvm/kvmd/commit/3ab43edeb962e95402ab0e18c553665ffd6117f0

PR #67, opened 2021-09-21
PWAY 16-port SH1601A Driver

SpookHCK:


mdevaev:
Sup. Why two modules?

kyounger:
I can't comment on the correctness of the code here, but I will point out that I also have an 8-port PWAY with the RS232 port and it also works with the same serial interface/protocol, so it might be worth designating this driver somehow as the 16-port flavor, (or maybe if there is some way to get an actual model number) instead of just "pway"?

mdevaev:
The driver that is located in the root directory of the kvm d module is not used by the system in any way. It is not clear from the PR which of the modules is correct. If they are intended for a different number of ports, then you need to make a parameter in get_plugin_options().

gitneko:
> I can't comment on the correctness of the code here, but I will point out that I also have an 8-port PWAY with the RS232 port and it also works with the same serial interface/protocol, so it might be worth designating this driver somehow as the 16-port flavor, (or maybe if there is some way to get an actual model number) instead of just "pway"?

If both 8 and 16 port PWAY use the same protocol, why should it be renamed? It should reather have a configurable port count option, so the same driver can be used for both port versions (so that it can error if a higher port identifier is given).

Or if possible, somehow find that out through the protocol, but if not you can only make a configurable option with the default value being the lowest port count (I don't know how the device reacts if you say to switch to port 14 and it only has 8 ports, someone would need to test that).

kyounger:
Oh, yes, I suppose if there is no conflict. I did notice the ports were enumerated in the code, but didn't know what was possible. Apologies for solutioning where I should have just been stating a use case.

SpookHCK:
Should only be one pway.py... I cleaned up part of the code then updated it...

mdevaev:
@SpookHCK Still two files.

mdevaev:
@SpookHCK ping

mdevaev:
@SpookHCK ping

SpookHCK:
Ok fixed the stupid Git thing with 2 files... should be only one new file in the correct location...

mdevaev:
Thanks. I'll fix the rest. It would be great if you added documentation for this switch here:
- https://github.com/pikvm/pikvm/blob/master/docs/multiport.md
- https://github.com/pikvm/pikvm/blob/master/docs/gpio.md

mdevaev:
Released in 3.49

SpookHCK:
The docs are in progress, just need a little more coffee and time :)

mdevaev:
Got it, thank you

kyounger:
Bump, @SpookHCK :)

PR #68, opened 2021-09-22
Option for use of create_subprocess_shell in run_process

aggieNick02:
Add an option to streamer configuration named 'use_shell' which allows us to use create_subprocess_shell instead of create_subprocess_exec in run_process.

This is useful when we want to run a streamer cmd that involves piping the output of one program to another program - e.g., using ffmpeg or gstreamer in conjunction with another program to get h264 frames to janus via memsink/etc.

There are also changes to kill_process, as otherwise the only program that would be killed when use_shell=True would be the top-level shell, with the children process continuing to run. I tried to preserve the behavior/logic for when there is only a single process to be the exact same, though I'll admit I didn't 100% follow the logic.

### DIFF
--- a/kvmd/aioproc.py
+++ b/kvmd/aioproc.py
@@ -21,6 +21,7 @@
 
 
 import os
+import signal
 import asyncio
 import asyncio.subprocess
 import logging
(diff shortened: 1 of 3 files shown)


aggieNick02:
To help motivate the "why", one nice capability this opens for me is the ability to very easily change between the janus ustreamer plugin and the janus streaming plugin. This is really useful when trying to determine if an issue I'm seeing may be on the ustreamer plugin side or the source h264 side. All I have to do is change janus config, tweak `stream.js` slightly, then take whatever program is writing to the memsink and have it pipe the frames to stdout instead. Then do:

`./my_streamer --use_stdout | ffmpeg -nostats -f h264 -i pipe: -an -c:v copy -f rtp rtp://127.0.0.1:8004`

mdevaev:
If you are doing something so custom, then why not just run a shell script that contains everything you need? At some point, the shell one-line will not be enough, and you will have to do it anyway, and this code will useless.

Actually, I don't like the idea of allowing shell execution in such places, because any escaping error or some bug can lead to complex problems.

aggieNick02:
That's a fair point about using a shell script. I'll be honest, I did not realize I could just toss my commands into a .sh file with `#!/bin/bash` at the top and get the same effect.

If I set the streamer command to be a shell script though, I still have the issue with `kill_process` only killing the top-level bash process and not the processes spawned by the script. To deal with those would require changes like the ones I made in `kill_process` . What are your thoughts on a PR with just the changes to `kill_process`?

mdevaev:
The lower-level application must correctly process SIGTERM and terminate its child processes (in shell, it is correct to do this using trap). Imagine that a streamer works at the top level that uses several processes and monitors them itself. SIGTERM will go to all processes, the children will fall and the monitoring process will try to do something about it.

But I think that sending all the processes in the SIGKILL group is a good idea just to make sure that there is nothing superfluous left.

mdevaev:
https://github.com/pikvm/kvmd/commit/487476afe48029e81e5acc14e01c162df0386050

aggieNick02:
Ah... I'm very much a linux newbie, so thank you for the patience and for explaining that. (Just went and read a quick tutorial on trap and signal forwarding and see how to do it pretty easily now.)

I also like the commit you just added - definitely seems like a good idea.

mdevaev:
Welcome :)

PR #70, opened 2021-10-03
only show keyboard message on homepage when relevant

patrickkettner:
this will hide the message when the app is running in the standalone mode being recommended


mdevaev:
Thanks!

PR #71, opened 2021-10-04
allow loading user customized css styles

ly4096x:
This will leave a 404 error in the console but it's safe to ignore. Also it's able to preserve the `user_custom.css` file across updates.


mdevaev:
Hi :) This feature is already exists:
- https://github.com/pikvm/kvmd/blob/master/configs/kvmd/web.css#L1
- https://github.com/pikvm/kvmd/blob/6b07a80834879970639b651b1a225f8d1f5f7c2e/configs/nginx/kvmd.ctx-server.conf#L36
- https://github.com/pikvm/kvmd/blob/master/web/base.pug#L33
- https://github.com/pikvm/kvmd/blob/master/web/index.html#L38

PR #72, opened 2021-10-04
web: move `user.css` to the last to allow overriding

ly4096x:


mdevaev:
The idea is good, but the implementation is wrong. I did the right thing: https://github.com/pikvm/kvmd/commit/1897e4d0434fa49ad97fc0448b9b8610b8c046ce

PR #79, opened 2022-01-28
Fix ocr cropping

mfunkey:


mdevaev:
I fixed it in a different way so as not to cause a memory leak with new image: de842f12e20d44de41d2649fb583064e958a2caf. Also my version passes all linters. I am grateful I left a comment in the commit :)

PR #80, opened 2022-03-11
Check that a packet start with PROTO::MAGIC

McNova:


mdevaev:
Why we need this?

McNova:
Because uart communication isn't packet and there is no warranty that client receive everything. If we miss some byte at the beginning then the client start receiving at an offset, after 8 bytes it will be sent to handlerequest and rejected because checksum don't match. And we'll get another offseted 8 bytes and reject again and again. I discovered that trying to port the protocol to micropyhton.

mdevaev:
I see, thanks. Now I need to review full protocol logic on the both sides. Jfyi.

mdevaev:
Sorry for the long answer. If a byte was lost, a timeout should be triggered and synchronization should occur. Have you encountered a problem for which this patch is needed?

mdevaev:
I will accept this PR, but I will comment on it, because I would like to get information about how you stumbled upon this and why timeout did not solve the problem.

PR #83, opened 2022-03-26
Add radius authentication support 

AndrewAubury:
Example usage here.

```
internal:
    force_users: admin
external:
    type: radiussrv
    hostsrv: 10.0.0.150
    port: 1812
    secret: PiKVM1 #KVM ID
```


kvmd/plugins/auth/radiussrv.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+    async def authorize(self, user: str, passwd: str) -> bool:
+        assert user == user.strip()
+        assert user
+        try:
+            r = radius.Radius(self.__secret, host=self.__hostsrv, port=self.__port)
+            return r.authenticate(user,  passwd)
+        except:
+            return False
```
mdevaev, line 70: No newline

kvmd/plugins/auth/radiussrv.py
```diff
@@ -0,0 +1,70 @@
...
+            "hostsrv":     Option("localhost"),
+            "port":  Option(1812),
+            "secret":  Option(""),
+            "user":    Option(""),
+            "passwd":  Option(""),
+            "timeout": Option(5.0),
+        }
+
+    async def authorize(self, user: str, passwd: str) -> bool:
+        assert user == user.strip()
```
mdevaev, line 64: Don't assert here for no good reason, in the normal working it should only return `True` or `False`. Use validators instead it.

kvmd/plugins/auth/radiussrv.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @classmethod
+    def get_plugin_options(cls) -> Dict:
+        return {
+            "hostsrv":     Option("localhost"),
+            "port":  Option(1812),
+            "secret":  Option(""),
+            "user":    Option(""),
+            "passwd":  Option(""),
+            "timeout": Option(5.0),
```
mdevaev, line 60: `type=valid_float_f01`

kvmd/plugins/auth/radiussrv.py
```diff
@@ -0,0 +1,70 @@
...
+        self.__port = port
+        self.__secret = secret
+        self.__user = user
+        self.__passwd = passwd
+        self.__timeout = timeout
+
+    @classmethod
+    def get_plugin_options(cls) -> Dict:
+        return {
+            "hostsrv":     Option("localhost"),
```
mdevaev, line 55: Rename hostsrv to host for the consistency with other modules. Also use `type=valid_ip_or_host`.

kvmd/plugins/auth/radiussrv.py
AndrewAubury: Changed this to a int, using valid_int_f1

PR #86, opened 2022-04-27
fix video pixel-imperfection

VoeGalore:
fix video pixel-imperfection by taking out 1px worth of internal padding, which inadvertently causing video stream resolution to swell at default(native) resolution

fixes [https://github.com/pikvm/pikvm/issues/599](url)

### DIFF
--- a/web/share/css/kvm/stream.css
+++ b/web/share/css/kvm/stream.css
@@ -55,7 +55,7 @@ div#stream-box {
 	object-fit: contain;
 	position: relative;
 	display: inline-block;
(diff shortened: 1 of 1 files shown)


VoeGalore:
I tested on multiple target resolution, tested chrome, firefox, edge(webkit) under Windows 10, tested chrome, chromium, ungoogle-chromium, firefox under ubuntu 20.04, it's working as intended.

mdevaev:
Okay, at least it doesn't make it worse

PR #88, opened 2022-05-19
Fix bug for some device after b4aeb86

xe5700:


mdevaev:
Sorry, for *which* device? This is not a bug, you should use the latest kernel from the pikvm repo.

xe5700:
I use rk3328 tv box uses old kernel 4.4.194 it not has these feature.

mdevaev:
Okay

PR #89, opened 2022-05-27
Support for driving TESmart KVMs by RS-232

TripleEmcoder:
Getting started:
- disable Linux serial console and debugging in /boot/cmdline.txt as mentioned in https://docs.pikvm.org/v3/#io-ports-and-jumpers in the orange box under UART
- prepare a console to TESmart cable, it's basically ethernet pins 4,5 from the middle of RJ-45 to the TESmart connector middle, and TXD, RXD lines according to any RJ45 Cisco cable pinout diagram (and the reverse after the first attempt fails)
- https://docs.pikvm.org/tesmart/#adding-ui-elements-to-control-the-kvm-switch - add new parameters to the driver:
```yaml
mode: 2 #RS-232
device: /dev/ttyAMA0
speed: 9600
```

Things to note (or change, feedback appreciated):

1. Changed pin numbers to go from 1 to 16 (for the largest model). The TESmart protocol, both TCP/IP and RS-232, and labels on the device itself, all use 1-based numbering. It really doesn't make sense to move that to 0..15 for PiKVM's config files. But that's just my opinion.

2. As noted in the comment, code requires serial_asyncio module. It's not yet available as an Arch package, so needs to be built from AUR. But you guys already build your own packages so I'm sure this can be handled. With serial_asyncio the only real change is how you open the connection, all read/write calls remained unchanged. This is greate for maintainability. Going the Ezcoo route would require a lot more changes.

3. I didn't find any examples in the other plugins on how to make driver parameters optional, so for now you need to specify both network and serial configuration and switch with a `mode` parameter. Which takes values 1 and 2, because I also couldn't find an example on how to validate a closed list of values like `[ "network", "serial" ]`.

### DIFF
--- a/kvmd/plugins/ugpio/tesmart.py
+++ b/kvmd/plugins/ugpio/tesmart.py
@@ -102,13 +102,14 @@ class Plugin(BaseUserGpioDriver):  # pylint: disable=too-many-instance-attribute
 
     @classmethod
     def get_pin_validator(cls) -> Callable[[Any], Any]:
-        return functools.partial(valid_number, min=1, max=16, name="TESmart channel")
+        return functools.partial(valid_number, min=0, max=15, name="TESmart channel")
 
     async def run(self) -> None:
         prev_active = -2
         while True:
             try:
-                self.__active = int(await self.__send_command(b"\x10\x00"))+1
+                # Current active port command uses 0-based numbering (0x00->PC1...0x0F->PC16)
+                self.__active = int(await self.__send_command(b"\x10\x00"))
             except Exception:
                 pass
             if self.__active != prev_active:
@@ -123,7 +124,8 @@ class Plugin(BaseUserGpioDriver):  # pylint: disable=too-many-instance-attribute
         return (self.__active == int(pin))
 
     async def write(self, pin: str, state: bool) -> None:
-        channel = int(pin)
+        # Switch input source command uses 1-based numbering (0x01->PC1...0x10->PC16)
+        channel = int(pin)+1
         assert 1 <= channel <= 16
         if state:
             await self.__send_command("{:c}{:c}".format(1, channel).encode())
--- a/kvmd/plugins/ugpio/tesmart.py
+++ b/kvmd/plugins/ugpio/tesmart.py
@@ -21,6 +21,12 @@
 
 
 import asyncio
+
+# At present this requires building a package from AUR:
+# https://aur.archlinux.org/packages/python-pyserial-asyncio
+# https://wiki.archlinux.org/title/Arch_User_Repository#Installing_and_upgrading_packages
+import serial_asyncio
+
 import functools
 
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


mdevaev:
Sup.

> Changed pin numbers to go from 1 to 16 (for the largest model).

All gpio plugins have pin numbering starting from 0, and this one should not be an exception for reasons of uniformity.

> asyncio_serial ... Going the Ezcoo route would require a lot more changes.

Acceptable.

> `mode` parameter

We can make exclusive settings. Like, if the serial port is specified, then use it, otherwise use tcp. It's a bit tricky, so I'll do it myself after merge

mdevaev:
Sup?

TripleEmcoder:
I will get back to this at the end of the month. If you really want pins to go from 0 I'll change that back and test in our environment. I'm going away for 2 weeks soon and have a lot of other stuff to finish. Sorry for the delay.

mdevaev:
Sup?

TripleEmcoder:
Done, thanks for the reminder and sorry for the delay.

mdevaev:
Thanks! I will refactor it a little.

mdevaev:
Okay, done. I removed `mode` argument, now if you specify `device` it will work over serial, but with `host` it will work over the network.

May I ask you to edit the documentation page https://docs.pikvm.org/tesmart? It's here: https://github.com/pikvm/pikvm/blob/master/docs/tesmart.md

TripleEmcoder:
> May I ask you to edit the documentation page https://docs.pikvm.org/tesmart? It's here: https://github.com/pikvm/pikvm/blob/master/docs/tesmart.md
Sure, I'd love to, but cannot make any promises as to when that will happen...

mdevaev:
Okay!

shimco:
Thanks for implementing this!
It would be great to get the documentation updated on this though, had been trying to achieve this setup for about 18 months now on multiple 16-port Tesmart KVM switches with the v3 Pikvms. Finally found mention of this via discussions in the discord and finally found this page which helped me actually get it working. 
A quick workaround for someone not using ATX control (most people using a multiport kvm) would be to use the included cat5 to ATX board with dupont cables and a cat5 cable to connect. The dupont cables will fit into the adaptor on the Tesmart, at least well enough for testing. 

/etc/kvmd/override.yaml
```
kvmd:
      gpio:
          drivers:
              tes:
                  type: tesmart
                  mode: 2 
                  device: /dev/ttyAMA0
                  speed: 9600
      scheme:
    
```

I would assume that serial would be a much better option for most people using this switch. Tesmart has recently done a youtube video series guide on setting things up working with Pikvm but has failed to include this option!
The only downside I've noticed so far is that I now don't have an indicator of which PC I'm on via the pikvm webgui.

Happy to contribute to getting better documentation for this if there's anything I can do, such as testing? ( I have access to multiple  8 and 16-port Tesmart kVM's and 4 pikvm's.

mdevaev:
Hello. Please make PR with your notes, with detailed descriptions. I don't use these switches, so I can make a mistake somewhere.

PR #91, opened 2022-07-06
add target all so build can be easly checked

tomaszduda23:


mdevaev:
Only one target can be assembled at a time, as this leaves an `.current` file artifact. If you want to do this just for testing, then let's do something like this:
```Makefile
_build:
        ...
_build_all: serial spi aum
        rm -f .current
```

tomaszduda23:
done

mdevaev:
Perfect

tomaszduda23:

PR #93, opened 2022-07-06
Change name to simplify interface getLedsAs->getLeds.

tomaszduda23:
It break abstraction between driver and protocol.
The driver type (win98 fix) is part of protocol anyway so I decided to do it that way.
Driver dependent part of PROTO could be put in separate file.
It would make PROTO less clean so I don't want to do that.

Part of https://github.com/pikvm/kvmd/pull/90

### DIFF
--- a/hid/lib/drivers/keyboard.h
+++ b/hid/lib/drivers/keyboard.h
@@ -1,0 +1,29 @@
+/*****************************************************************************
+#                                                                            #
+#    KVMD - The main PiKVM daemon.                                           #
+#                                                                            #
+#    Copyright (C) 2018-2022  Maxim Devaev <mdevaev@gmail.com>               #
+#                                                                            #
+#    This program is free software: you can redistribute it and/or modify    #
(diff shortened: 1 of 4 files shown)


mdevaev:
I propose to make a structure containing the state of the LEDs, initialize it in _sendResponse(), pass a pointer to it in getLeds() and add it to main.cpp something like conversion function.

Kinda `keyboard.h`:
```cpp
typedef struct {
        bool caps;
        bool scroll;
        bool num;
} LedsState;

hids:
```cpp
void getLeds(LedsState *state);
```

`main.cpp`:
```cpp
uint8_t _ledsToProto(LedsState *state) {
        uint8_t leds = 0;
        if (state->caps) leds |= PROTO::PONG::CAPS;
        if (state->scroll) leds |= PROTO::PONG::SCROLL;
        if (state->num) leds |= PROTO::PONG::NUM;
}
```

mdevaev:
I've refactored the code a little (coding style, some minor things), please take a look.

PR #100, opened 2022-07-09
Stm32 init

tomaszduda23:


tomaszduda23:
What was the reason to replace setup/loop with main?

mdevaev:
> What was the reason to replace setup/loop with main?

USB must be initialized after the HID configuration, otherwise the report descriptor will contain unnecessary elements.

mdevaev:
Before this patch, I have to test all the changes.

mdevaev:
The keyboard is completely broken. The mouse works very slowly, with huge delays. These issues need to be fixed before we move on.

mdevaev:
Okay, false alarm, there are some problems with the driver on the host. But I think we should ask users to test the new firmware.

tomaszduda23:
I also checked with gdb serial hid and it looks the same. More testing is of course good idea.

mdevaev:
1. Can we use `main()` for both platforms? You will probably need this to configure USB on STM32.
2. Please send all AVR changes as a separate commit if it's not too much trouble.

tomaszduda23:
1. That is an option. To remove USBDevice.attach(); is just one line. stm32 does not use anything like that. I would rather path //USBDevice.attach(); in avr main and use setup/loop. It will be easier to port for other platform and easier to read by everyone.
2. OK
3. I can do that.
> break the building of all types of env except AUM

what do you have in mind?

mdevaev:
We have three avr hid configurations: serial, spi and aum. They require the same libraries, but are built with different flags. You have moved the libraries from the general section to aum.

tomaszduda23:
https://github.com/pikvm/kvmd/pull/100/checks it is pass. _atmelavr was added for common libs. 
```
	git+https://github.com/Harvie/ps2dev#v0.0.3
	digitalWriteFast@1.0.0
```
looks like only for aum. 

Anyway I is good idea to separate ini files.

tomaszduda23:
Ahhh. You are right. I didn't consider code which is commented out.

What about adding setup/loop and changing main in avr.

mdevaev:
Okay

tomaszduda23:
https://github.com/pikvm/kvmd/pull/102

mdevaev:
So, now about this thing. Have you made all the changes you need in the basic firmware structure to make STM32? I'd rather take this as a completely done job (but little STM32-related stuff is still in separate commits).

tomaszduda23:
It seems like I have all I need. I was going to commit device by device. Since it will be easier to review. I can implement serial, keyboard, mouse. I don't think that I would do spi, win98, ps2.

tomaszduda23:
Anyway I can make serial, keyboard, mouse in one commit.

mdevaev:
Okay. In fact, win98 is a pretty simple fix, it just requires a shift to the left for coordinates, look at the patch: https://github.com/pikvm/kvmd/blob/master/hid/patches/hid-win98.patch

mdevaev:
Sorry, I missed your message. For example, your host may be finicky and want to work only with the keyboard, and due to the presence of a mouse, it may be buggy. That is, this is done for paranoid reasons, and the need to control the mouse and keyboard is a feature of the protocol.

tomaszduda23:
This is pretty much all. You can check status at https://github.com/tomaszduda23/kvmd/tree/stm32_init/hid/lib/drivers-stm32

mdevaev:
Let's try to get rid of `keymap.cpp` we have [keymaps.csv](https://github.com/pikvm/kvmd/blob/master/keymap.csv) which is used by [genmap.py](https://github.com/pikvm/kvmd/blob/master/genmap.py) to generate all the keymaps in the repository at once to keep them consistent.
So:
1. Remove `keymap.cpp`
2. Create `keymap.h.mako` [like this](https://github.com/pikvm/kvmd/blob/master/hid/lib/drivers-avr/usb/keymap.h.mako) using `otg_key` (USB HID hex codes) instead of `arduino_name`.
3. Add the new mako file to the [root Makefile](https://github.com/pikvm/kvmd/blob/master/Makefile#L205).
4. Run `make regen` and check if everything will work fine.

hid/patches/platformio-stm32f1-no-serial-usb.patch
```diff
@@ -0,0 +1,11 @@
+--- aaa/tools/platformio-build-stm32f1.py	2022-07-16 18:54:42.536695468 +0200
```
mdevaev, line 1: Wrong newlines in this file

hid/platformio-stm32.ini
```diff
@@ -0,0 +1,65 @@
...
+extra_scripts =
+	post:patch.py
+
+[_common]
+lib_deps =
+	git+https://github.com/ZulNs/STM32F1_RTC#v1.1.0
+	git+https://github.com/arpruss/USBComposite_stm32f1#3c58f97eb006ee9cd1fb4fd55ac4faeeaead0974
+	drivers-stm32
+build_flags =
+# ----- The default config with dynamic switching -----
```
mdevaev, line 19: Remove not implemented flags from examples please

hid/lib/drivers-stm32/usb/mouse-absolute-stm32.h
```diff
@@ -0,0 +1,89 @@
...
+			void sendWheel(int delta_y) override {
+				_mouse.move(0, 0, delta_y);
+			}
+
+			bool isOffline() override {
+				return USBComposite == false;
+			}
+
+			void periodic() override {
+#if 0
```
mdevaev, line 73: Is it really needed? If not, then delete it

hid/lib/drivers-stm32/usb/keyboard-stm32.h
```diff
@@ -0,0 +1,127 @@
...
+				if (usb_code != KEY_ERROR_UNDEFINED) {
+					if(offset) {
+						usb_code += KEY_HID_OFFSET;
+					}
+					state ? _keyboard.press(usb_code) : _keyboard.release(usb_code);
+				}
+			}
+
+			void periodic() override {
+#if 0
```
mdevaev, line 95: Why is that?

hid/patches/platformio-stm32f1-no-serial-usb.patch
tomaszduda23: This was done on purpose. Destination file has CRLF so without it patch was not working.

mdevaev: This is a a python code, why it's not working? Please show the error

hid/lib/drivers-stm32/usb/mouse-absolute-stm32.h
tomaszduda23: I removed debug code. I just keep method in interface to keep consistency between keyboard and mouse.

hid/lib/drivers-stm32/usb/keyboard-stm32.h
tomaszduda23: I removed it.

hid/lib/drivers-stm32/usb/mouse-absolute-stm32.h
mdevaev: No problem with `periodic()`

hid/patches/platformio-stm32f1-no-serial-usb.patch
tomaszduda23: https://unix.stackexchange.com/questions/239364/how-to-fix-hunk-1-failed-at-1-different-line-endings-message

Hunk #1 FAILED at 1 (different line endings). <- otherwise I'm getting this kind of error

mdevaev:
Do I understand correctly that you used HID-Project library?

By the way, we can communicate a little faster, my discord is open.

tomaszduda23:
For me it is side project. I check it when I have spare time so github communication is fine.

I used that one https://github.com/arpruss/USBComposite_stm32f1
and key codes from https://github.com/NicoHood/HID/blob/master/src/KeyboardLayouts/ImprovedKeylayouts.h

HID-Project does not compile on stm32
https://github.com/NicoHood/HID/issues/195
https://github.com/NicoHood/HID/issues/287
I didn't want to dig into those issues so just found another lib.

hid/patches/platformio-stm32f1-no-serial-usb.patch
mdevaev: AH this is a patch, I'm stupid. Okay.

mdevaev:
Okay, no problem.

> key codes from https://github.com/NicoHood/HID/blob/master/src/KeyboardLayouts/ImprovedKeylayouts.h

I got rid of these constants in arduino, you can just use my mako template. The constants are the same because it is a USB HID table.

mdevaev:
That's not what I meant. Please do a rebase, I threw out the arduino_code altogether so that unified usb codes could be used.

hid/.gitignore
```diff
@@ -2,3 +2,4 @@
 /.current
 /.vscode/
 /.config
+/platformio.ini
```
mdevaev, line 5: Why?

mdevaev:
Okay, let me merge this and try to fix it myself, it seems I've completely confused you.

tomaszduda23:
btw this lib assume different values for shift etc https://github.com/arpruss/USBComposite_stm32f1/blob/3c58f97eb006ee9cd1fb4fd55ac4faeeaead0974/Keyboard.cpp#L181

PR #101, opened 2022-07-10
docker proof of concept

tomaszduda23:
This is WIP base on old code. @mdevaev please let me know if you could consider to merge it. Comments are welcome.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,114 @@
+FROM python:3.9-alpine3.14 as builder
+RUN apk add --no-cache \
+		alpine-sdk \
+		linux-headers \
+		libjpeg-turbo-dev \
+		libevent-dev \
+		libbsd-dev \
+		autoconf \
+		automake \
+		autoconf-archive \
+		libtool
+
+COPY patches /patches/
+RUN git clone --depth 1 --branch v$(cat /etc/alpine-release) git://git.alpinelinux.org/aports && \
+	cd aports/main/musl && \
+	abuild -F unpack && \
+	sed -i '/pthread_setname_np/a int pthread_getname_np(pthread_t, char *, size_t);' ./src/v1.2.2/include/pthread.h && \
+	cp /patches/pthread_getname_np.c src/v1.2.2/src/thread && \
+	patch -d src/v1.2.2 -p1 < /patches/getnameinfo.patch && \
+	abuild -F build && \
+	cd src/v1.2.2 && \
+	make install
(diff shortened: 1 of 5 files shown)


mdevaev:
The `configs` directory contains files for production, so they cannot be changed for these purposes.

And also what you wanted to do has already been implemented :) See `testenv` directory and `Makefile` in the root of the repo. You can use `make run` to run kvmd + web ui.

tomaszduda23:
> The `configs` directory contains files for production, so they cannot be changed for these purposes.

I was going to add override.
> And also what you wanted to do has already been implemented :) See testenv directory and Makefile in the root of the repo. You can use make run to run kvmd + web ui.

I saw that but it is not exactly the same thing. The goal here would be to do as much as possible during building and starting container so it would be easy to use. Other services could be running inside as well. I just added kvmd + web ui since it was minimum for PoC.

mdevaev:
I'm not sure. The idea is that kvmd is not bound to a specific platform and it doesn't matter how you run it.

mdevaev:
I meant that now the components work as separate microservices and are tested that way. This is quite convenient because it allows me to restart a separate service, rather than the entire container. Plus, I use Arch inside the container, just like in a live PiKVM OS, to bring the test conditions closer to real ones.

tomaszduda23:
> This is quite convenient because it allows me to restart a separate service, rather than the entire container. 

There are 2 ways to solve this:
a) homeassistant has one supervisor container which create other services inside new containers. It is quite complex and requires exposing docker socket inside container.
b) use supervisord or similar solution so you can control services without restarting container

> Plus, I use Arch inside the container, just like in a live PiKVM OS, to bring the test conditions closer to real ones.

I also though about Arch. The official image seems not support arm or arm64. I though about using something which allows to use systemd inside containers but it also seems to be overkill.
I ended up using alpine since it is very small. It is also not perfect since I had to patch during image building.

mdevaev:
To be honest, I don't see any need for this. The test environment is working fine now. There are things we can't test right now (journald and janus gateway), and if you want to help with the testing environment, it's better to focus on that.

tomaszduda23:
My goal here was to make docker image which could be used instead of kvm os. I put there testenv dir only to solve temporary issue.

mdevaev:
I need to think about it. Right now I don't know why this may be needed in upstream.

tomaszduda23:
Let's close it for now. It would be better to base it on ustreamer docker image.

PR #107, opened 2022-08-13
add SW reset

tomaszduda23:
reset PIN on kvmd is optional. HID works without problem until _reset_required is set. After that device has to be restarted manually. With SW reset it is possible to switch between HID without reset PIN.


mdevaev:
Sup. I'm busy with other things for now, so I'll review your patches a little later.

mdevaev:
Refactored a little also.

PR #112, opened 2022-08-16
add the ability to use command exit status as a gpio state in ugpio/cmd

Luberry:
I wanted to be able to use an arbitrary commands output as a gpio state. This makes it easy to script arbitrary devices/statuses for display in the ui.

### DIFF
--- a/kvmd/plugins/ugpio/cmd.py
+++ b/kvmd/plugins/ugpio/cmd.py
@@ -63,7 +63,7 @@ class Plugin(BaseUserGpioDriver):  # pylint: disable=too-many-instance-attribute
 
     @classmethod
     def get_modes(cls) -> Set[str]:
-        return set([UserGpioModes.OUTPUT])
+        return set([UserGpioModes.OUTPUT, UserGpioModes.INPUT])
 
     @classmethod
     def get_pin_validator(cls) -> Callable[[Any], Any]:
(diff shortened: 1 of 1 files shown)


mdevaev:
:ok_hand:

mdevaev:
released

mdevaev:
I have separated this functionality into a separate module `cmdret`, since the `read` operation is performed every time there are changes in any gpio plugin.

Luberry:
> I have separated this functionality into a separate module `cmdret`, since the `read` operation is performed every time there are changes in any gpio plugin.

Could you have not just returned false on read if the mode was output? Then it would skip calling the command unless it was an input.

mdevaev:
I don't understand :/

Luberry:
> I don't understand :/

Essentially. Since im assuming your issue is that the command will be run twice when a gpio is set to output. Instead of fully splitting this into its own thing. You would do something like (pseudocode becYse I don't have the actual code in front of me on my phone)

```
//Inside read function
if this.mode != Input {
    return false
}

//Insert code to run and read command output here

```

Basically only do anything in the read function if the gpio is set as an input.

mdevaev:
Okay, now I understand :ok_hand: To be honest, I don't see a fundamental difference between mode and a separate module. In my opinion, a separate module is even more convenient, you will definitely not forget to set the necessary mode.

Luberry:
> Okay, now I understand ok_hand To be honest, I don't see a fundamental difference between mode and a separate module. In my opinion, a separate module is even more convenient, you will definitely not forget to set the required mode.

Yeah, if it were me personally i would err on the side of keeping one module since you are emulating gpio, and it would make sense to use one plugin to have full gpio functionality for any given thing. But that's just my personal opinion, and as i didn't create the project i will defer to your judgement as the maintainer.

PR #113, opened 2022-08-29
Zoom and Enhance before running OCR

adamoutler:
Before analyzing the OCR text, Greyscale the image, zoom it to 2x size, and run a bicubic interpolation (4x4 anti-aliasing for upsampling).  This makes it easier for Tesseract to read the text.

### DIFF
--- a/kvmd/apps/kvmd/tesseract.py
+++ b/kvmd/apps/kvmd/tesseract.py
@@ -41,6 +41,7 @@ from typing import Set
 from typing import Generator
 from typing import Optional
 
+from PIL import ImageOps
 from PIL import Image as PilImage
 
 from ...errors import OperationError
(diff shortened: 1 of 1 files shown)


adamoutler:
Tests show very positive results.  I was able to copy a few long URLs without problem. Whereas before 1-out-of-10 chars was wrong, it is now more like 1-out-of-1000. On one instance out of 10 attempt, an `&` symbol was converted to `8`.   

Future improvements may include increasing from 2x to 3x scaling, and adding a ` image.filter(ImageFilter.SHARPEN)`.  This would require testing for both OCR performance and User Experience for extra processing time.  I'm operating on the actual hardware so further experimentation isn't practical for me.

mdevaev:
Perfect, thank you!

PR #114, opened 2022-09-25
Allow to run with unpatched kernel

hzyitc:
These day, I'm trying to port pikvm to other devices. I notice that it can run well with unpatched kernel. So I think we can keep this compatibility.


mdevaev:
It will not work at all. This flag exists for a reason and my patch with it was even accepted into the upstream kernel. Without it, it is impossible to unlock and eject the mass storage device on some OS like Linux.

PR #115, opened 2022-09-25
Use `dbus` instand of `dbus-next`

hzyitc:
These day, I'm trying to package `kvmd` as debian package. So we can support more devices easily with [`Armbian`](https://github.com/armbian/build). But unfortunately, Debian don't have package - `python3-dbus-next`.

I'm not sure whether you like this change. Or only apply this patch durning build debian package?

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -51,7 +51,6 @@ depends=(
 	python-netifaces
 	python-systemd
 	python-dbus
-	python-dbus-next
 	python-pygments
 	python-pyghmi
 	python-pam
--- a/kvmd/apps/kvmd/sysunit.py
+++ b/kvmd/apps/kvmd/sysunit.py
@@ -22,38 +22,33 @@
 
 import types
 
-import dbus_next
-import dbus_next.aio
-import dbus_next.aio.proxy_object
-import dbus_next.introspection
-import dbus_next.errors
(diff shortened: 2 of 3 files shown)


mdevaev:
This code should be asynchronous and fast, so dbus-next is used here, and the regular dbus library in this project is deprecated at all. I'm sorry, but I can't accept this and I prefer that the dependency problem be solved by adding packages to the distribution.

hzyitc:
Fine. But at this time, could you accept to use patch (store in `debian/` directory) to address this question?
Or still add this package into depends and let user download the debs.

[hzyitc/kvmd-debian](https://github.com/hzyitc/kvmd-debian) is my repo to pack kvmd as debian package.

I have tested it on Armbian with my arm board.

mdevaev:
This patch is incorrect. You can't just replace one library with another, because what you're doing may cause the main KVMD loop to hang. At some point if systemd responds slowly you will get problems.

hzyitc:
Sorry. I check the debian package again. The `python3-dbus-next` was added 2 month ago, but only for `bookworm` and `sid`. Ubuntu still doesn't have this package. So I think I can continue to pack it. Thanks for your reply.

mdevaev:
:ok_hand:

PR #116, opened 2022-10-18
nginx: serve ACME challenges over unencrypted HTTP

woju:
This PR makes nginx serve ACME challenges from port 80, without redirecting to HTTPS. Redirect to HTTPS is pointless (HTTPS doesn't add anything).

It also prevents obtaining keys for machines without public IP, but connected over wireguard (or other VPN) to a machine with public IP, and with public DNS pointing to the public proxy, not to the real pikvm in the intranet. For me this is a preferred mechanism since it avoids disseminating DNS keys to all machines.

### DIFF
--- a/configs/nginx/nginx.conf
+++ b/configs/nginx/nginx.conf
@@ -39,6 +39,7 @@ http {
 	server {
 		server_name localhost;
 		include /etc/kvmd/nginx/listen-http.conf;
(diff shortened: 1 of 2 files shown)


mdevaev:
It makes sense, thank you. You can also add the doc PR to pikvm/pikvm repo.

woju:
Here you are: pikvm/pikvm#834.

Thanks!

PR #119, opened 2023-02-16
Add gzip support to nginx.conf

michael-sayapin:
When the Web UI is used over constrained connections, gzip compression might help make it snappier. At the default compression level it consumes a negligible amount of CPU resources.

### DIFF
--- a/configs/nginx/nginx.conf
+++ b/configs/nginx/nginx.conf
@@ -27,6 +27,10 @@ http {
 	keepalive_timeout 10;
 	client_max_body_size 4k;
 
+	gzip                    on;
+	gzip_proxied            any;
+	gzip_types              text/plain text/css application/json text/javascript;
+
 	client_body_temp_path	/tmp/kvmd-nginx/client_body_temp;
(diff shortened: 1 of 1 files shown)


mdevaev:
I've tried it. Without compression: 767Kb, with gzip: 565Kb. Not a big deal. What do you think?

mdevaev:
Sup?

NiKiZe:
IMHO Compression takes almost no CPU these days, there is no reason to not have it enabled, even if the reduced size is minimal.

mdevaev:
If this does not give a win, there is no point in warming the CPU and increasing the entropy of the universe :)

mdevaev:
But seriously speaking, 200 kilobytes at the start is something, since then we will have a video working. The only thing that would make sense is the compression of websocket traffic, but I don't see Nginx working with this.

https://datatracker.ietf.org/doc/html/rfc7692

mdevaev:
Given the potential for a BREACH attack and almost zero compression profits, I don't think it makes sense to enable gzip.

PR #121, opened 2023-03-12
adding ch9329 HID support

jacobbar:
the ch9329 is a plug and play solution for HID emulation through serial, it supports absolute mouse, relative mouse and a keyboard


mdevaev:
I will merge it into the separate branch, fix some things and merge to the master.

PR #125, opened 2023-03-29
Correct copy behavior during temporary directory set up

vipergts450:
Corrects a bug in which authentication files stored in the `/runroot` directory, rather than subdirectories, would be missed while the script sets up the temporary working directory for certificate renewals. This leads to a failure of the rest of the renewal procedure in kvmd-certbot due to it being unable to find the authentication file after switching to the temporary working directory. This patch changes the copy behavior to copy all files and subdirectories rather than just the 3 subfolders of `/runroot`.

### DIFF
--- a/scripts/kvmd-certbot
+++ b/scripts/kvmd-certbot
@@ -108,7 +108,7 @@ case "$1" in
 	renew)
 		shift
 		create_tmp
-		cp -a "$cur"/{config,work,logs} "$tmp"
+		cp -a "$cur"/* "$tmp"
 		chown -R "$user:" "$tmp"
 		sed -s -i -e "s| = $cur/| = $tmp/|g" "$tmp/config/renewal/"*
 		sudo --preserve-env -u "$user" certbot renew "$@" \


vipergts450 (commented):
Corrects a bug in which authentication files stored in the /runroot directory, rather than subdirectories, would be missed while the script sets up the temporary working directory for certificate renewals. This leads to a failure of the rest of the renewal procedure in kvmd-certbot due to it being unable to find the authentication file after switching to the temporary working directory. This patch changes the copy behavior to copy all files and subdirectories rather than just the 3 subfolders of /runroot.

jeholliday:
@vipergts450 It seems like your solution fixed the issue observed in #980 for Google DNS, but I am still running into this issue when using Cloudflare DNS. I think that root cause is that the bash glob `"$cur"/* ` does not match hidden files in `$cur/`. The instructions in the handbook say to name the file `.cloudflare.auth`, and this file isn't being copied. I'd suspect that anyone using Route53 would run into the same problem because the instructions say to name the file `.route53.auth`.

I found that I could work around this issue by renaming the file to `cloudflare.auth`, updating the renewal config file to use the new file, and rerunning the renewal.
```
kvmd-pstrun -- mv /var/lib/kvmd/pst/data/certbot/runroot/.cloudflare.auth /var/lib/kvmd/pst/data/certbot/runroot/cloudflare.auth
kvmd-pstrun -- nano /var/lib/kvmd/pst/data/certbot/runroot/config/renewal/<domain>.conf
kvmd-certbot renew --force-renewal
```

This issue could either be resolved by updating the handbook instructions or by updating the glob pattern to include hidden files.

vipergts450:
My Google auth file actually is a hidden auth file (`.googledns.auth` in my
case) also and in my testing was copied by `cp -a`. Once you changed your
filename to not be a dotfile it worked? Weird.

vipergts450:
@jeholliday You're right, there's something else amiss here. In bash, `cp -a <dir>/*` indeed skips dotfiles. I will amend and resubmit a pull using `cp -a <dir>/.` instead.

jeholliday:
Great, thanks for taking a look at this. The syntax for ` cp -a <dir>/*` is confusing because it's being parsed separately by bash and cp.

mdevaev:
Merged

PR #127, opened 2023-04-04
Update kvmd-certbot: fix cp -a to catch dotfiles

vipergts450:
Second pull request for this bug. Apparently, in bash `cp -a <dir>/*` ignores dotfiles, which are often how authfiles get named, notably as per the documentation for SSL cert management for certbot plugins in the PiKVM Wiki. Using `cp -a <dir>/.` instead should catch dotfiles and all files and subdirectories correctly.


mdevaev:
Thank you!

PR #128, opened 2023-04-12
Remove unused items

adamoutler:
These items are called out as inactive or removed from pylint.


adamoutler:
I see we are using pylint==2.10.2.  The current default version is pylint 2.17.2.  The version difference is causing various failures.   The newer version claims all of these items are unused and has additional checks.  Will continue to try to resolve.  https://hackedyour.info/Nj16GDAZmAkYVijc

PR #132, opened 2023-04-26
Set focus on text area when text is clicked.

adamoutler:
This patch removes one of the 4-steps required in order to copy and paste into PiKVM.   When the Text menu is clicked, the focus is set on the Text Area.


mdevaev:
https://github.com/pikvm/kvmd/commit/3170dcdb575124a6867381b09af6a2a18a6b1278

PR #133, opened 2023-05-12
Merge strategy expansion

adamoutler:
This commit allows `Merge Strategy:` to be applied within the root of any YAML document loaded by KVMD. The strategies are as follows:

- MERGE :  The default strategy.  It is similar to a recursive "override" behavior of all dictionaries and values.
- DEEP_MERGE (aka "deep merge" or "deepmerge"): will merge dictionaries and will append lists.
- APPEND: Appends new values to dictionaries, adds new values to lists, but ignores existing non-dictionary keys.
- DISABLED (aka "disable"): will not merge content in the YAML.

This enables granular control over YAML files, enabling a simple "download" of configurations to be applied to KVMD, allowing more flexibility to fix problems or add features rapidly.  For rapid prototyping, configurations can be enabled/disabled quickly.  For test environments, a configuration can be applied and patched via a selected merge strategy. 

The overall logic flow can be shown as follows: 
image

I conducted tests with 2-files the "Mouse Jiggler"(90-jiggler)  and "Second Mouse Jiggler"(91-jiggler).  The changes can be seen here in the following pictures.

Default, no strategy or `MERGE` strategy:
image

The `DEEP MERGE` strategy:
image

The `DISALBLED` strategy:
image

Assuming we have a base YAML of `/etc/kvmd/override.d/0-view-table`
``` YAML
kvmd:
    gpio:
        view:
            table:
                - ["#This is a displayed message in the table."]
```
Under normal operations, the following YAML will clobber the message in the table resulting in total replacement.  `/etc/kvmd/override.d/91-message`
``` YAML
kvmd:
    gpio:
        view:
            table:
                - ["#This message replaces the message in the table."]
```
If we wish to add to this table, the keyword `Merge Strategy:` must be used in conjunction with the `APPEND` or `DEEPMERGE` strategies.  The strategies themselves are case-insensitive. A `DEEP_MERGE` Strategy can be applied as follows  `/etc/kvmd/override.d/91-message`
``` YAML
Merge Strategy: Deep Merge
kvmd:
    gpio:
        view:
            table:
                - ["#This message is also in the table"]
```
The primary difference between `APPEND` and `DEEPMERGE` strategy is the `DEEPMERGE` strategy will replace values within python `dict`, whereas `APPEND` takes a lighter approach and will only affect items which are not already defined.

For the `DISABLED` strategy, it makes things easier for removing without deleting or commenting out sections. We can disable the `91-message` by applying `DISABLED` strategy.
``` YAML
Merge Strategy: disable
kvmd:
    gpio:
        view:
            table:
                - ["#This will not appear in the table because it is disabled"]
```

Future improvements: Implement the tracked file parameter.  Currently the file being operated upon is passed as a parameter while merge operations are being conducted, but adding it to configuration causes problems with drivers.  So while it is present, it is currently unused except for debug purposes.

Recommended user technical documentation:

# Applying Merge Strategies to your YAML files in KVMD
When using PiKVM, you might often have multiple YAML configuration files for different modules or functionalities. There might be instances where these configurations need to be combined or "merged" to form a unified configuration. The PiKVM offers a range of merge strategies to handle such scenarios, enabling fine-grained control over the merging process to control the system configuration.

## Understanding Merge Strategies
A merge strategy defines how values from one configuration (source) are combined with existing values in another configuration (destination). In this context, the destination is the system configuration, and after the merge, your YAML file becomes a part of the system configuration. The following merge strategies are available in KVMD:

**Merge (default):** New values from the source file overwrite existing values in the destination file. If the value is a list, the entire list from the source file replaces the list in the KVMD configuration.

**Append:** New values are appended to existing lists in the destination file. New keys from the source file are added to the KVMD configuration, but existing values for common keys are not overwritten.

**Deep Merge:** New values from the source file overwrite existing values in the destination file. If the value is a list containing dictionaries, the specific dictionaries from the source file are merged with the dictionaries in the KVMD configuration, matching on dictionary keys.

**Disabled:** The merge operation is skipped entirely.

## Specifying the Merge Strategy in your YAML
To specify a merge strategy, add a key called `Merge Strategy` at the root level of your source YAML file, with the value set to the desired strategy. The merge strategy is case-insensitive and accepts spaces and underscores. For example, all of the following are valid ways to specify the deep merge strategy:

``` yaml
Merge Strategy: deep_merge
Merge Strategy: Deep Merge
Merge Strategy: DEEPMERGE
```
If no `Merge Strategy` is specified, the default is `Merge`.

## Example
Consider you have the following KVMD configuration:

``` yaml
a: "1"
b: "2"
c: [1, 2, 3, {"d": 4}]
```
And have added this to your Override.d:

``` yaml
Merge Strategy: append
a: "3"
b: "2"
c: [3, 4, 5, {"e": 6}]
```
After placing both files in the KVMD configuration directory and restarting the KVMD service, the resulting configuration would look like this:

``` yaml
a: "1"
b: "2"
c: [1, 2, 3, {"d": 4},  4, 5, {"e": 6}]
```
Note how the values for `a` and `b` from the source file are ignored, and the list `c` is appended with the new values.

## How to Apply
Simply place your configuration files into the appropriate directory (typically `/etc/kvmd/override.d`) and specify your merge strategy as shown above. After you have done this, reboot the unit, or restart the `kvmd` service for the changes to take effect:

``` bash
sudo systemctl restart kvmd.service
```
The KVMD service will automatically merge the configurations based on the specified merge strategies.

We hope this guide helps you in effectively managing and merging your YAML files. For further assistance or queries, refer to the PiKVM documentation or reach out to the support team.


adamoutler:
Will fix mypy problems.  I wasn't checking them.

adamoutler:
The goal behind this PR was to provide a way for users to add features such as MJ in a modular manner.  Since there appears to be no use case for it now, I'm closing the PR.

PR #134, opened 2023-05-14
Show first time boot status in oled

srepac:


scripts/kvmd-bootconfig
```diff
@@ -82,6 +89,10 @@ if [ -n "$FIRSTBOOT$FIRST_BOOT" ]; then
 		unset disk part npart label
 	fi
 
+    ### create NFS_Primary directory for use with NFS MSD after reboot ###
```
mdevaev, line 92: This is not related with the current PR, pelase remove this part

scripts/kvmd-bootconfig
```diff
@@ -48,6 +48,13 @@ rw
 # ========== First boot configuration ==========
 
 if [ -n "$FIRSTBOOT$FIRST_BOOT" ]; then
+    ### stop regular kvmd-oled service and show first time setup status in oled
+    systemctl stop kvmd-oled
+    /usr/bin/kvmd-oled --interval=0 --text="First time setup\nDo NOT interrupt\nPlease Wait...\n"
+    
+    ### force reboot after first boot
```
mdevaev, line 55: Why?

scripts/kvmd-bootconfig
```diff
@@ -48,6 +48,13 @@ rw
 # ========== First boot configuration ==========
 
 if [ -n "$FIRSTBOOT$FIRST_BOOT" ]; then
+    ### stop regular kvmd-oled service and show first time setup status in oled
+    systemctl stop kvmd-oled
```
mdevaev, line 52: Why we need to stop it here?

scripts/kvmd-bootconfig
srepac: If you don't temporarily stop kvmd-oled service, the message will get clobbered.

scripts/kvmd-bootconfig
srepac: ok

scripts/kvmd-bootconfig
srepac: If you don't temporarily stop kvmd-oled service, the message will get clobbered.

scripts/kvmd-bootconfig
mdevaev: I mean why you need a reboot here?

scripts/kvmd-bootconfig
mdevaev: Got it

PR #135, opened 2023-05-23
sysctl: tune dirty data writeback policies

arter97:
The existing policy sets dirty_background_bytes to 5, which is erroneously too small that incurs write amplification and read starvations. Considering a raspi with 2 GiB of RAM, this is 43 million times smaller than the default Linux policy, which is 10%.

Considering that this will be deployed to mostly read-only system with emphasis on data integrity (hence needing tighter data writeback), set this to 256 KiB.

Continuing addressing the overly excessive data writeback policies, also bump the dirty_writeback_centisecs from 250ms to 1000ms. Any dirty data written in the past 1 second will be written to the storage.

### DIFF
--- a/configs/os/sysctl.conf
+++ b/configs/os/sysctl.conf
@@ -1,6 +1,7 @@
-# https://github.com/raspberrypi/linux/issues/1753
-vm.dirty_background_bytes = 5
-vm.dirty_writeback_centisecs = 25
+# Overly-aggressive data writeback policies to minimize the chance of data corruption
+# Start dirty data writeback when it exceeds 256 KiB or 1s has passed
+vm.dirty_background_bytes = 262144
+vm.dirty_writeback_centisecs = 100
 
 # https://groups.google.com/g/meetecho-janus/c/xoWIQfaoJm8
 net.core.rmem_default = 500000


adamoutler:
Considering this is a system which generally uses RO mode except for a single filesystem, sometimes, what impact might a large dirty buffer have on system operations?  Is it possible that instead of immediate denial of write due to full buffer, the system may cache a relatively large portion of data in RAM to be written?   

I think overall this may improve RW access to the MSD partition in the rare event of small file transfer, but I'm curious to know what effect this may have when the system is operating in its typical RO mode.

arter97:
> Considering this is a system which generally uses RO mode except for a single filesystem, sometimes, what impact might a large dirty buffer have on system operations?

Large dirty writeback buffer just means write will be delayed for that much longer. This boosts general I/O performance: read will be faster as writes don't block read requests as much, write will be faster as it's cached more aggressively in RAM.

This is not something that we'd want much from a system that's mission-critical and want data to be kept intact as much as possible, even in the events of a power loss.

Also, by "Considering that this will be deployed to mostly read-only system", I meant that write performance is not as important. This policy doesn't make any sense for systems that frequently writes data like a general Linux desktop.

> Is it possible that instead of immediate denial of write due to full buffer, the system may cache a relatively large portion of data in RAM to be written?

Denial of write doesn't happen due to dirty writeback buffer, but yes, the system may cache more portion in RAM, and that's bad in perspective of data integrity as it increases the chance of more data getting lost in the events of a power loss.

> I think overall this may improve RW access to the MSD partition in the rare event of small file transfer, but I'm curious to know what effect this may have when the system is operating in its typical RO mode.

This is not just for MSD. Small file transfer can occur a lot in the rootfs file-system, and the most prominent example I can think of is the use of the package manager - pacman. The value of 5 will slow down its operation by a lot with no meaningful protection improvements over something like 256K, which is also already really low.

mdevaev:
The old parameters were used for paranoid reasons :) I tested the MSD download speed and didn't see much difference in behavior, so I think nothing will work worse. Thanks for the patch.

PR #138, opened 2023-06-23
Add secure cookie params

ssza:
This pull request adds httponly, secure and samesite cookie parameters.


mdevaev:
Hello. Why we need it? Is this not a warning fix for Firefox?

mdevaev:
I've added `httponly=True, samesite="Strict"` without `secure=True` because we have use case with disabled HTTPS.

PR #141, opened 2023-08-16
Add config items to allow changing inquiry string of otg msd drive

gy6221:
As subject.


mdevaev:
Thanks for the idea, but I made alternative configuration structure and a kernel patch for this.

PR #149, opened 2023-09-16
pico hid: better ps2x2pico integration

No0ne:


mdevaev:
This is it! Thank you!

PR #155, opened 2024-01-23
Add Unifi GPIO driver

appkins:
Adds a GPIO driver for Unifi devices that can control POE state.

- pikvm/pikvm#1212

Configuration for the driver is as follows:

```yaml
    kvmd:
        gpio:
            drivers:
                unifi:
                    type: unifi
                    host: https://10.0.0.1:443
                    user: admin
                    passwd: admin
                    mac: ff:ff:ff:ff:ff:f1
                    verify: false
            scheme:
                server1_power:
                    driver: unifi
                    pin: 0
                    mode: output
                # cycle mode will quickly cycle a port's power state
                server1_cycle:
                    driver: unifi
                    pin: 0:cycle
                    mode: output
                    switch: false
            view:
                table:
                    - ["Unifi Controller"]
                    - []
                    - ["#Server 1", "server1_power", "server1_cycle|Reset"]
    ```

### DIFF
--- a/kvmd/plugins/ugpio/unifi.py
+++ b/kvmd/plugins/ugpio/unifi.py
@@ -1,0 +1,373 @@
+# ========================================================================== #
+#                                                                            #
+#    KVMD - The main PiKVM daemon.                                           #
+#                                                                            #
+#    Copyright (C) 2018-2023  Maxim Devaev <mdevaev@gmail.com>               #
+#                                                                            #
+#    This program is free software: you can redistribute it and/or modify    #
+#    it under the terms of the GNU General Public License as published by    #
+#    the Free Software Foundation, either version 3 of the License, or       #
+#    (at your option) any later version.                                     #
+#                                                                            #
+#    This program is distributed in the hope that it will be useful,         #
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of          #
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the           #
+#    GNU General Public License for more details.                            #
+#                                                                            #
+#    You should have received a copy of the GNU General Public License       #
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.  #
+#                                                                            #
+# ========================================================================== #
+
+
+import asyncio
+import functools
+import json
+
+from typing import Callable
+from typing import Any
+
+from multidict import CIMultiDict, CIMultiDictProxy
+
+import aiohttp
+
+from ...logging import get_logger
+
+from ... import tools
+from ... import aiotools
+from ... import htclient
+
+from ...yamlconf import Option
+
+from ...validators.basic import valid_stripped_string_not_empty
+from ...validators.basic import valid_bool
+from ...validators.basic import valid_float_f01
+
+from . import BaseUserGpioDriver
+from . import GpioDriverOfflineError
+
+
+# =====
+
+def set_poe_mode(pin: str, state: bool, port_override: dict[str, Any]) -> dict[str, Any]:
+    port_idx = str(port_override["port_idx"])
+    if port_idx == pin:
+        port_override["poe_mode"] = "auto" if state else "off"
+    return port_override
+
+
+# pylint: disable=too-many-instance-attributes disable=too-many-arguments
+class Plugin(BaseUserGpioDriver):
+    def __init__(
+        self,
+        instance_name: str,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mdevaev** [comment]
Please fix linter errors

**Comment by appkins** [comment]
> Please fix linter errors

It should be ready for another test. I've fixed all the flake8 errors.

**Comment by mdevaev** [comment]
Mypy again

**Comment by appkins** [comment]
> Mypy again

I fixed all the mypy errors. Had to refactor some of the typings, should be good for another test.

**Comment by mdevaev** [comment]
Let me know when you consider it ready.

**Comment by appkins** [comment]
I made a few revisions after continued testing. The driver supports `cycle` mode due to the reset time being significantly faster than turning the POE mode on/off.

The current implementation looks for a suffix in the pin name. `{i}:cycle` is the current pattern. I'm happy to make this configurable. Having multiple unifi drivers seems to cause a bunch of issues for both the toggle mode and the cycle mode.

**Comment by appkins** [comment]
It's ready. Here's a demo I recorded to show the functionality.

https://github.com/pikvm/kvmd/assets/13772502/4f617618-9da7-42a7-9347-601bdbe0d165

**Comment by mdevaev** [comment]
I'm going to merge it into a separate branch, make some refactoring and merge it into master.

**Comment by mdevaev** [comment]
Late question but anyway. Why you're using string validator and the check for numeric for pins? And no numeric check for input. Is pin a number or not?

```python
    def get_pin_validator(cls) -> Callable[[Any], Any]:
        return valid_stripped_string_not_empty

    def register_input(self, pin: str, debounce: float) -> None:
        _ = debounce
        self.__state[pin] = None

    def register_output(self, pin: str, initial: (bool | None)) -> None:
        if pin.isnumeric():
            self.__initial[pin] = initial
            self.__state[pin] = None

    def prepare(self) -> None:
        for (pin, state) in self.__initial.items():
            if pin.isnumeric():
                if state is not None:
                    self.__state[pin] = state
```

mdevaev:
Also please test my branch https://github.com/pikvm/kvmd/tree/unifi-gpio

I rewrote the logic of working with HTTP and removed the duplicate code, -80 lines. If everything is OK, then I will continue, I see more ways to make it easier.

mdevaev:
I don't quite understand the processes inside the main loop and how the data is processed. Can you describe to me which structures are stored where and what they contain? All these tables look like they should be combined into one structure with dataclasses. Please describe them here.

mdevaev:
@appkins sup?

appkins:
> Late question but anyway. Why you're using string validator and the check for numeric for pins? And no numeric check for input. Is pin a number or not?
> 
> ```python
>     def get_pin_validator(cls) -> Callable[[Any], Any]:
>         return valid_stripped_string_not_empty
> 
>     def register_input(self, pin: str, debounce: float) -> None:
>         _ = debounce
>         self.__state[pin] = None
> 
>     def register_output(self, pin: str, initial: (bool | None)) -> None:
>         if pin.isnumeric():
>             self.__initial[pin] = initial
>             self.__state[pin] = None
> 
>     def prepare(self) -> None:
>         for (pin, state) in self.__initial.items():
>             if pin.isnumeric():
>                 if state is not None:
>                     self.__state[pin] = state
> ```

Hey, sorry for the delay. I’ve been very busy the past few weeks. 

This exists to support single instance states for switch type outputs, while also supporting additional outputs for the same port for cycling.

Since state is a dict, I opted to exclude any duplicate ports with the ‘:cycle’ suffix. Otherwise we’d see a collision in the hash map.

Since the initial values are ignored anyway, I think it’s safe to omit the init for state.

appkins:
> I don't quite understand the processes inside the main loop and how the data is processed. It's quite difficult to understand and support. Can you describe to me which structures are stored where and what they contain? All these tables look like they should be combined into one structure with dataclasses. Please describe them here.

The UniFi api is a pain to work with. Omitting a set value effectively removes it from the internal UniFi state for a port. This will do things like reset the attached network of a port etc. To work around this problem, it’s necessary to pass all values of the UniFi port table into the update request.

I think a struct would be preferable. Though it will need two structures. One for the port map, another for the port table.

I’ll supply some more detail on this shortly.

mdevaev:
Sup?

mdevaev:
Sup?

mdevaev:
@appkins are you still interested in this?

mdevaev:
Sup?

PR #157, opened 2024-01-28
Keyboard Issues #19 fix (https://github.com/No0ne/ps2x2pico/issues/19)

No0ne:
This fixes two bugs regarding ghost mouse movement and key presses mentioned here: https://github.com/No0ne/ps2x2pico/issues/19

The default 10ms polling rate could be a bit too much for some systems.
Thats why I still need to implement the mouse decoupling mentioned here: https://github.com/No0ne/ps2x2pico/issues/1#issuecomment-1849435280


mdevaev:
Thank you sir!

PR #158, opened 2024-01-31
Add plugin for XH-HK4401 clone with different control codes

lumascet:
I bought apparently a slightly different version of the XH-HK4401 4 port kvm. It uses a shorter code for switching outputs and does not report back the switched channel if it is set via the control port. it will only report if one physical button is pressed. I altered the python code of the xh_hk4401.py and created a new h4401.py (i just made this name up, i dont think it has a model number). Now my kvm is working as indended.

Here is the store link of the kvm: https://www.aliexpress.com/item/1005005979139402.html

### DIFF
--- a/kvmd/plugins/ugpio/h4401.py
+++ b/kvmd/plugins/ugpio/h4401.py
@@ -1,0 +1,198 @@
+# ========================================================================== #
+#                                                                            #
+#    KVMD - The main PiKVM daemon.                                           #
+#                                                                            #
+#    Copyright (C) 2018-2023  Maxim Devaev <mdevaev@gmail.com>               #
+#                  2021-2021  Sebastian Goscik <sebastian.goscik@live.co.uk> #
+#                                                                            #
+#    This program is free software: you can redistribute it and/or modify    #
+#    it under the terms of the GNU General Public License as published by    #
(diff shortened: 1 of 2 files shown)


mdevaev:
Hello. Please find some kind of marking on the case or inside on the pcb, I would not like to call the module by a random name

lumascet:
Good point. I opened up the kvm and there are two boards with different labels.

- the HDMI switch board has the label `HK4501-V0-A1-HDMI`
- the USB/control board has the label `HK4401-V1.9-B-USB`

I think it would make more sense to use the usb board name, but thats exactly the same. Maybe the new product name is HK4501 and they just reused the older control board? What do you think?

IMG_4287
IMG_4288

mdevaev:
Probably some kind of variation. Okay, I'll just add support for this protocol to the existing xk module.

mdevaev:
Please test `xh_hk4401.py` from the master branch and set option `protocol: 2` in the gpio config for it.

lumascet:
Thanks! Changes working as intended.

mdevaev:
:ok_hand:

mdevaev:
I will make a new release with it now.

PR #160, opened 2024-02-17
ps2: keyboard passthru support

No0ne:
I'm currently implementing this: https://github.com/pikvm/pikvm/issues/476

At the moment only keyboard passthru works.
The GPIOs used are 26=data and 27=clock.

@mdevaev please try it out, if you are happy I'll continue with mouse support.

### DIFF
--- a/hid/pico/src/ph_ps2.c
+++ b/hid/pico/src/ph_ps2.c
@@ -31,7 +31,8 @@
 #define _LS_POWER_PIN	13
 #define _KBD_DATA_PIN	11 // CLK == 12
 #define _MOUSE_DATA_PIN	14 // CLK == 15
-
+#define _KBD_PT_DATA_PIN	26 // passthru, CLK == 27
+#define _MOUSE_PT_DATA_PIN	16 // passthru, CLK == 17
 
 u8 ph_g_ps2_kbd_leds = 0;
 bool ph_g_ps2_kbd_online = 0;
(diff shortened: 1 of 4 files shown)


mdevaev:
Thanks for that! I don't have a PS/2 keyboard right now, I'll buy one soon and test it.

hid/pico/src/ph_ps2.c
```diff
@@ -123,13 +126,13 @@ void ph_ps2_mouse_send_wheel(s8 h, s8 v) {
 
 void ph_ps2_send_clear(void) {
 	if (PH_O_IS_KBD_PS2) {
-		for(u8 key = 0xe0; key <= 0xe7; key++) {
-			kb_send_key(key, false, 0);
-		}
+		//for(u8 key = 0xe0; key <= 0xe7; key++) {
```
mdevaev, line 129: BTW why you commented this?

No0ne: This doesn't work. The queue is only 8 bytes long, so blasting it with circa 200 bytes of key-up events won't do it. I'll have to think about key tracking so only those keys which where held down get reset with key-up events.

mdevaev: Got it.

mdevaev:
Sup! I finally found PS/2 keyboard and will test it today or tomorrow.

mdevaev:
Sometimes it skips some key presses. Perhaps because CLK sometimes ends in different ones.

RigolDS5
RigolDS6
RigolDS7

No0ne:
Ok I'll continue on that. I did a larger refactor, please make a new **ps2 passthru branch** in the kvmd repo and I'll make a new pull request.
But first I'll get the main branch onto the pre-1.0 release which has major mouse support fixes.

mdevaev:
:ok_hand:

PR #161, opened 2024-03-11
On bad network conditions `disconnected` can happen periodically

RSATom:
but it doesn't mean channel is already broken.
It's better wait `failed` state since `disconnected` can be changed to `connected` at any moment.
Fixes pikvm/pikvm#1084
Also fixes reconnect on Firefox.


RSATom:
Also this patch requires https://github.com/meetecho/janus-gateway/pull/3342 to work as expected.

PR #162, opened 2024-03-15
Add plugin for Extron USB and SW VGA switchers

nvllsvm:
This adds support for various Extron USB and VGA switchers (possible other Extron switchers given how both USB and VGA share the same serial API).

I was hesitant to open this PR due to being niche, but figured it's a good place for conversation.


mdevaev:
Thank you! Please make some documentation here: https://docs.pikvm.org/gpio/

It's here: https://github.com/pikvm/pikvm/blob/master/docs/gpio.md

PR #166, opened 2024-05-17
Introduce multi type HID

JeWe37:
This introduces the ability to have different HID plugins acting as the keyboard and mouse respectively via a `multi` plugin that dispatches to the correct HID plugin. 

This properly resolves pikvm/pikvm#1291


mdevaev:
Sup. I like the approach, but I don't like magic inside otg plugin to support multihid. I can take this into a separate branch and redo something according to my understanding of how it should be, and you will test the result. What do you say?

JeWe37:
Sure I don't mind, go ahead. I won't be able to test for the next few days though, but will get back to this after.

mdevaev:
Good. Okay, I'll run linters on your code, please fix all errors and warning before merging.

JeWe37:
I think that should be all the linter issues fixed.

mdevaev:
Let's rerun checks then.

JeWe37:
Looks fine except for one issue in pylint, but I didn't even touch the function it's complaining about?

mdevaev:
It's a false-positive in the new pylint release. I've fixed it.

mdevaev:
JFYI I don't forget about it. Sorry for the delay, I have a very priority task right now.

PR #168, opened 2024-06-02
Update kvmd-gencert to include IP and unique hostname

adamoutler:
This change is required to provide a Subject Alternate Name for the certificate rather than no SANs at all.  The default "localhost" is retained.  New certified IPs/hostnames include
- pikvm-${ProcessorSerialNumber}.local
- Ethernet IPv4 Address
- Ethernet IPv6 Address
- WiFi IPv4 Address (if connected)
- WiFi IPv6 Address (if connected)

This change allows a user to trust the PiKVM cert as a Root cert by default. After generating a cert with the `scripts/kvmd-gencert`, any acquired IPv4 and IPv6 addresses along with `pikvm-${ProcessorSerialNumber}` are registered as subject alternate names for the cert.   This allows the certificate to be used as a Trusted Root Certificate.

In Windows/Chrome
1.  show the certificate
image
2. Click Details, then Export then save to your Desktop.
image
3. On the desktop double-click the file and then "Install Certificate..."
image
4. Choose Current User, then next
image
5. Choose Place all certificates in the following store, then Browse...
image
6. Press Trusted Root Certification Authorities, then Ok, and then Next
image
7. Press Finish
image
8. Press Yes, then Ok
image
9. Close your browser completely, then reopen to your PiKVM.  The certificate is installed
image

Possible Future improvements:
- make the IP or `pikvm-${ProcessorSerialNumber}` the official cert name instead of localhost. This will help to identify the certificates in the keystore.
- Create a certificate regeneration extras which runs this cert-genetator and restarts the webserver so that the user can recertify a dynamic IP address on-the-fly.

### DIFF
--- a/kvmd/apps/__init__.py
+++ b/kvmd/apps/__init__.py
@@ -179,7 +179,7 @@ def _init_config(config_path: str, override_options: list[str], **load_flags: bo
     scheme = _get_config_scheme()
     try:
         yaml_merge(raw_config, (raw_config.pop("override", {}) or {}))
-        yaml_merge(raw_config, build_raw_from_options(override_options), "raw command line options")
+        yaml_merge(raw_config, build_raw_from_options(override_options), "raw CLI options")
         _patch_raw(raw_config)
         config = make_config(raw_config, scheme)
 
--- a/kvmd/yamlconf/merger.py
+++ b/kvmd/yamlconf/merger.py
@@ -19,10 +19,25 @@
 #                                                                            #
 # ========================================================================== #
 
-from typing import Optional
-
 
 # =====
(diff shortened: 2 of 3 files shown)


scripts/kvmd-gencert
```diff
@@ -20,44 +20,71 @@
...
 # ========================================================================== #
 
-
 set -e
 export LC_ALL=C
 
 if [ "$(whoami)" != root ]; then
-	echo "Only root can do that"
-	exit 1
+    echo "Only root can do that"
```
mdevaev, line 27: Please don't change tabs to spaces

adamoutler: Fixed

PR #170, opened 2024-06-08
Enable Avahi

adamoutler:
### Summary

This PR introduces support for Avahi ZeroConf. With this implementation, the PiKVM will show up in Nautilus/Dolphin under Network or in Finder on Mac. 

### Details

The implementation allows for auto-discovery and unique identification of the PiKVM without relying on dynamic information such as MAC address, IP address, or HTTPS certificate. This feature is particularly useful for adding devices to Home Assistant.

### ZeroConf Data

The ZeroConf data broadcasts the following `_pikvm._tcp` registration:

Screenshot_20240608-142322

The same data is also replicated with the `_https._tcp` registration.

### Benefits

- **Auto-discovery**: Simplifies the process of finding PiKVM devices on the network.
- **Unique Identification**: Each PiKVM is uniquely identifiable without relying on changing network information.
- **Home Assistant Integration**: Facilitates the addition of PiKVM devices to Home Assistant, enhancing smart home automation capabilities.

### Script Flow

1. **Retrieve Serial and Model Numbers**: The script reads the serial number from `/proc/device-tree/serial-number` and the model number from `/proc/device-tree/model`.

2. **Generate Avahi Configuration**: The script defines the new Avahi service configuration, including HTTP and HTTPS services with associated TXT records for serial and model numbers.

3. **Compute Configuration Hashes**: The script computes the MD5 hash of the current and new configurations to determine if updates are necessary.

4. **Update Avahi Service File**: If the configuration has changed, the script writes the new configuration to the Avahi service file.

5. **Install and Start Avahi**: The script checks if Avahi is installed, installs it if necessary, and ensures the Avahi daemon is enabled and started, if not disabled by the user.


mdevaev:
I've made an alternative implementation: https://github.com/pikvm/kvmd/commit/5c3ac4c9c100fe381591f8756fca3bd56b9ca9f7

Just install and enable avahi service, and run trigger FIRSTBOOT=1.

mdevaev:
Okay, I added ENABLE_AVAHI to the boot config and after the update you will get avahi to the dependencies. This is still disabled by default, including for new images, so I don't want to change the default behavior, at least for now.

To get this working, after the update (not yet released in the release) you will need to create `/boot/pikvm.txt`:
```ini
ENABLE_AVAHI=1
```

You can do the same on fresh PiKVMs on their memory cards as part of the pre-deployment configuration. This will enable Avahi. I will consider enabling this by default in the future, but for now at least you have the option to use it.

PR #174, opened 2024-07-16
Proper manifest file for app mode

grantemsley:
Filling in the name and adding start_url will make browsers offer to install PiKVM as an application, including adding shortcuts to the start menu and everything. This runs it in app mode just like the instructions for running chrome with the --app flag.

Requires a valid SSL certificate be setup before browsers will show the option.


mdevaev:
Thank you!

PR #178, opened 2024-08-15
Support NETRC to download files

jlegaul7:
As it stands, kvmd does not use netrc to download files which makes downloading files off of site requiring authentication infeasible.

this commit adds the missing flag to let aiohttp use netrc for authetication


mdevaev:
Maybe settings in override.yaml will be a better solution?

mdevaev:
Sup?

mdevaev:
Probably it's not relevant anymore.

PR #179, opened 2024-08-26
configuration option to disable the Info API's 'hw' and 'fan' sections

czo:
On my ancient board too many things are missing (no device-tree, missing sensors, never had a fan etc). To mitigate device-tree and temperature readings and errors, this configuration option can be used to disable it:

kvmd:
    info:
        disabled:
          - hw
          - fan

The error messages:

kvmd.apps.kvmd.info.hw           ERROR --- Can't read DT model from /proc/device-tree/model: [Errno 2] No such file or directory: '/proc/device-tree/model'
kvmd.apps.kvmd.info.hw           ERROR --- Can't read DT serial-number from /proc/device-tree/serial-number: [Errno 2] No such file or directory: '/proc/device-tree/serial-number'
kvmd.apps.kvmd.info.hw           ERROR --- Can't read CPU temp from /sys/class/thermal/thermal_zone0/temp: [Errno 2] No such file or directory: '/sys/class/thermal/thermal_zone0/temp'

### DIFF
--- a/kvmd/apps/__init__.py
+++ b/kvmd/apps/__init__.py
@@ -393,6 +393,10 @@ def _get_config_scheme() -> dict:
                     "timeout":    Option(5.0, type=valid_float_f01),
                     "state_poll": Option(5.0, type=valid_float_f01),
                 },
+                "disabled": Option(
+                    [],
+                    type=functools.partial(valid_string_list), unpack_as="disabled",
+                ),
             },
 
             "log_reader": {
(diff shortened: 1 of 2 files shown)


mdevaev:
You don't need to disable `fan` because it doesn't hardly required. Just set `kvmd/info/fan/unix` to an empty string to disable it.

As for `hw`, removing this part breaks the API for the Web UI. I would recommend patching `hw.py` for your platform.

mdevaev:
Okay, I thought about it and it won't be accepted because it breaks the API. For your case you need to implement hw subinfo module.

PR #180, opened 2024-08-26
when a plugin name contains a dot, import it from the global space to…

czo:
Currently, the plugins (which is specified in the main.yaml) always be imported from the "kvmd.plugins." space. This small change enable to specify the full module path in the main.yaml to use off the tree code for 3rd party implementations.


mdevaev:
Reviewed it again. KVMD plugins are in the `kvmd.plugins` package and you can just install your own plugins there. This is one of the [recommended](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/) ways to manage plugins in Python.

PR #181, opened 2024-08-26
ability to pass the kvmd instance to the plugin, because this will al…

czo:
Because in my case the both the HID and the ATX "functions" are provided by the same Arduino device, some modifications had been required to enable function calls between those two plugins. The least invasive ide for this, is to store the parent kvmd instance in the plugin, and thereafter, the plugin can now access the kvmd and it's variables. Because the _hid are already existed, i created a corresponding _atx too.

### DIFF
--- a/kvmd/apps/kvmd/server.py
+++ b/kvmd/apps/kvmd/server.py
@@ -172,6 +172,7 @@ class KvmdServer(HttpServer):  # pylint: disable=too-many-arguments,too-many-ins
 
         self.__auth_manager = auth_manager
         self.__hid = hid
+        self.__atx = atx
         self.__streamer = streamer
         self.__snapshoter = snapshoter  # Not a component: No state or cleanup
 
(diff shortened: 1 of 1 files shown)


mdevaev:
I suppose one of the plugins does all the work, including communication with the device, while the second one just makes calls to the first one. Since plugins are not used independently or in other software, the best solution would be to make a global reference to the initialized object inside the first plugin, then import it from the second plugin. Like this:

fancy_hid_plugin.py:
```python
HidObject = None

class Plugin(...):
    def __init__(self, ...):
        ...
        HidObject = self
```

fancy_atx_plugin.py:
```python
import fancy_hid_plugin

print(fancy_hid_plugin.HidObject)
```

This way you can do without hacks in the base KVMD.

czo:
Thanks, i will check it in the next couple of days

czo:
This was a good idea, works perfectly.

mdevaev:
Nice.

PR #182, opened 2024-08-26
silence the systemd/dbus exception if there are no matching services

czo:
I got some strange dbus/systemd related exceptions when loading the landing page:

```
root                             ERROR --- add match request failed. match="sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',path='/org/freedesktop/DBus',member='NameOwnerChanged'", [Errno 32] Broken pipe
root                           WARNING --- a message handler threw an exception on shutdown
Traceback (most recent call last):
  File "/home/czo/opt/venv-312/pikvm/lib64/python3.12/site-packages/dbus_next/aio/message_bus.py", line 63, in write_callback
    self.offset += self.sock.send(self.buf[self.offset:])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/czo/opt/venv-312/pikvm/lib64/python3.12/site-packages/dbus_next/message_bus.py", line 457, in _finalize
    handler(None, err)
  File "/home/czo/opt/venv-312/pikvm/lib64/python3.12/site-packages/dbus_next/message_bus.py", line 586, in reply_notify
    callback(reply, err)
  File "/home/czo/opt/venv-312/pikvm/lib64/python3.12/site-packages/dbus_next/message_bus.py", line 970, in add_match_notify
    if msg.message_type == MessageType.ERROR:
       ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'message_type'

```

I figured out, this will happens if the MessageBus connection was opened, but there are no matching service was found on my system and the connection will be closed without any function call. Clearing up the kvmd/extras folder will be too invasive to me, so i figured out, this exception will did not pop up if any systemd related function was called. So, getting the default target will solves the problem. I put it into a try:except block, because i did not know if it can or can not thorw an exception. Maybe, this is a bug in dbus_next?

### DIFF
--- a/kvmd/apps/kvmd/sysunit.py
+++ b/kvmd/apps/kvmd/sysunit.py
@@ -75,6 +75,10 @@ class SystemdUnitInfo:
     async def close(self) -> None:
         try:
             if self.__bus is not None:
+                try:
+                    await self.__manager.call_get_default_target()
+                except:
+                    pass
                 self.__bus.disconnect()
                 await self.__bus.wait_for_disconnect()
         except Exception:


mdevaev:
Do you have more logs? I'm wondering what was going on above.

czo:
Umm, nope, but it's seems to be, not related to this project. I mean, this small snippet also crashing if I commenting out the "call_get_default_target". 

```
(pikvm) [czo@crux-linux ~]$ cat test.py 
import asyncio
import dbus_next
import dbus_next.aio

async def test():
    bus = await dbus_next.aio.MessageBus(bus_type=dbus_next.BusType.SYSTEM).connect()
    intr = await bus.introspect("org.freedesktop.systemd1", "/org/freedesktop/systemd1")
    systemd = bus.get_proxy_object("org.freedesktop.systemd1", "/org/freedesktop/systemd1", intr)
    manager = systemd.get_interface("org.freedesktop.systemd1.Manager")
    
    await manager.call_get_default_target()
    
    bus.disconnect()
    await bus.wait_for_disconnect()


asyncio.run(test())
(pikvm) [czo@crux-linux ~]$ 

```

mdevaev:
I see.

> await manager.call_get_default_target()

Where did you find this hack?

czo:
> Where did you find this hack?

Nowhere. I was figured out, when runned some tests, that closing the connection without any real request will causes a problem. I think, the default target will be exists in every systemd installation, this will be the most safest workaround to mitigate this exception.

mdevaev:
I'm just curious what exactly `call_get_default_target` does.

PR #183, opened 2024-08-28
Adds tapo smart plug GPIO Plugin

chriswiggins:
Brings in a tapo smart plug integration for GPIO. This supports devices as follows: https://www.tapo.com/en/product/smart-plug/

The only thing that this requires is the `tapo` python module, which I'm not sure how to include, or whether or not this project will be happy to have that included? Is there a way to specify this as an optional dependency? I've just been testing by running:

```sh
rw
pacman -S python-pip
pip3 install tapo --break-system-packages
ro
```

I looked into how this was done inside the python module, and its relatively complex so not just as simple as bringing in an HTTP call.

Example `/etc/kvmd/override.yaml`:

```yaml
kvmd:
  gpio:
    drivers:
      plug_1:
        type: tapo
        ip: 1.2.3.4
        email: your@email.domain
        password: my-super-secure-password

    scheme:
      plug_1_switch:
        driver: plug_1
        pin: 0
        mode: output
        switch: true
        initial: null
        pulse:
          delay: 0
    view:
      header:
        title: Power
      table:
        - ["#Power Switches"]
        - []
        - ["#Switch 1", plug_1_switch]
```

Please provide any and all feedback!

### DIFF
--- a/kvmd/plugins/ugpio/tapo.py
+++ b/kvmd/plugins/ugpio/tapo.py
@@ -1,0 +1,149 @@
+# ========================================================================== #
+#                                                                            #
+#    KVMD - The main PiKVM daemon.                                           #
+#                                                                            #
+#    Copyright (C) 2018-2024  Maxim Devaev <mdevaev@gmail.com>               #
+#                                                                            #
+#    This program is free software: you can redistribute it and/or modify    #
+#    it under the terms of the GNU General Public License as published by    #
+#    the Free Software Foundation, either version 3 of the License, or       #
+#    (at your option) any later version.                                     #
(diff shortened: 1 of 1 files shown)


mdevaev:
Hello. Do you really need entire rust library for this? This is not packed in distro, and we have no mechanism for optional deps. As I understand, tapo api is just a REST, maybe you can change the plugin to make http calls directly and avoid this dep?

chriswiggins:
Hey @mdevaev - very valid point, I did think about that but it initially seemed a bit complex. Let me go back to the drawing board and work out if there's another option 👍

PR #187, opened 2025-03-11
web: disable autocomplete for 2fa

akehir:
Since 2FA codes are always changing, it is not helpful to have autocomplete suggestions for this field. It just displays a long list of outdated codes.

With the html attribute `autocomplete` this behaviour can be disabled. In my opinion that is always desired for the 2fa input.


mdevaev:
Thanks!

PR #189, opened 2025-04-13
customise nginx listen host

ananthb:
Customise the IPv4 and IPv6 hosts that nginx listens on.


ananthb:
I forgot to add a few imports. I'll push a fix for it shortly.

ananthb:
@mdevaev it's fixed now.

mdevaev:
There are two problems:
* `[::]` is not allowed for IPv6 because it's using for host:port notation.
* I think we should split HTTP/HTTPS addresses additional.
I'll fix/implement it myself and mention in the commit that this is based on your patch.

PR #192, opened 2025-05-31
Fixed OLED geometry detection logic for non-CM4 devices

Maotechh:
On a Raspberry Pi Zero 2 W with a 128×64 OLED display, the screen was incorrectly initialized as 128×32 due to hardcoded logic relying solely on device model and USB presence. This caused display issues for non-CM4 boards with 64px-high screens.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 *.pyc
 *.swp
 /venv/
+.DS_Store
--- a/kvmd/apps/oled/__init__.py
+++ b/kvmd/apps/oled/__init__.py
@@ -44,6 +44,53 @@ _logger = logging.getLogger("oled")
 
 # =====
 def _detect_geometry() -> dict:
+    # First try to detect actual OLED device height by creating a temporary device
+    try:
+        from luma.core.interface.serial import i2c
+        from luma.oled.device import ssd1306, sh1106
+        
+        for address in [0x3C, 0x3D]: # Common I2C addresses
+            try:
+                serial = i2c(port=1, address=address)
+                # SSD1306
+                try:
+                    temp_device = ssd1306(serial)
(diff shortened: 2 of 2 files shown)


mdevaev:
Hello.

I don't want to accept it because of too much magic here. If you're using a very custom build, just create systemd override and pass the correct --width and --height parameters to the display service.

Maotechh:
Hi, thanks for the feedback.

I understand the concern about adding too much "magic" or implicit behavior. However, I’d like to clarify that this change doesn’t introduce unpredictable logic—it simply adds a lightweight and robust auto-detection fallback **only when the display parameters aren’t explicitly provided**.

Here’s why I think it’s justified:

1. **The current detection logic is already a heuristic (CM4 + USB → 64px)** and fails silently for many real-world cases like the Raspberry Pi Zero 2 W with a 128×64 OLED. So the existing code already contains "magic"—but it’s **fragile and incorrect** in certain scenarios.

2. **The proposed detection is non-intrusive**:

   * It only runs if no width/height is explicitly configured.
   * It tries to read the actual display height using existing drivers (SSD1306/SH1106).
   * It gracefully falls back to the current logic if detection fails.
     There’s **no added configuration burden** or required dependencies beyond what's already in use.

3. **Telling users to write a systemd override adds unnecessary complexity**, especially for less technical users. For most people using standard 128×64 OLEDs, this patch makes the software **just work**, which improves out-of-the-box experience.

4. **This change improves compatibility without breaking anything.** It retains full backward compatibility and existing behavior for known devices.

If you'd prefer, I can refactor the detection logic under a `try_detect_oled_height()` helper function to make it clearer and less "magical." But I do believe this behavior is practical and user-friendly, and it resolves a real hardware mismatch without requiring per-user hacks.

PR #209, opened 2026-04-09
web: allow server-side defaults for page-close and stream-suspend settings

nullstacked:
## Summary

The Web UI has ~18 client-side settings with defaults hardcoded in JavaScript. Only one of them (`Expand for the entire tab by default`) uses the existing `tools.config.getBool()` mechanism to allow server-side configuration via CSS custom properties in `/etc/kvmd/web.css`.

This PR extends that same pattern to **all** remaining settings, so admins can configure defaults for every UI toggle server-side.

## What changed

Every `tools.storage.bindSimpleSwitch()` call with a hardcoded default now reads from `tools.config.getBool()` (or `tools.config.get()` for non-boolean values) first, falling back to the original default. No behavior changes when no CSS properties are set.

**Files changed:**
- `web/share/js/kvm/main.js` — page close confirmation
- `web/share/js/kvm/stream.js` — stream suspend, mic, cam, mode, orientation
- `web/share/js/kvm/atx.js` — ATX confirmation dialog
- `web/share/js/kvm/switch.js` — switch ATX/MSD confirmations
- `web/share/js/kvm/hid.js` — SysRq confirmation
- `web/share/js/kvm/keyboard.js` — bad link, swap Ctrl/CapsLock
- `web/share/js/kvm/mouse.js` — squash, reverse scroll X/Y, cumulative scroll, dot cursor
- `web/share/js/kvm/paste.js` — paste confirmation, secure mode

## Example usage

Admins can now set defaults in `/etc/kvmd/web.css`:

```css
:root {
    /* Interface settings */
    --config-ui--kvm--page-close-ask: 0;
    --config-ui--kvm--full-tab-stream: 1;      /* already worked before */
    --config-ui--kvm--stream-suspend: 1;

    /* Stream settings */
    --config-ui--kvm--stream-mode: mjpeg;       /* janus, media, or mjpeg */
    --config-ui--kvm--stream-orient: 0;         /* 0, 90, 180, 270 */
    --config-ui--kvm--stream-mic: 0;
    --config-ui--kvm--stream-cam: 0;

    /* Confirmation dialogs */
    --config-ui--kvm--atx-ask: 0;
    --config-ui--kvm--switch-atx-ask: 0;
    --config-ui--kvm--switch-msd-ask: 0;
    --config-ui--kvm--hid-sysrq-ask: 0;
    --config-ui--kvm--hid-pak-ask: 0;
    --config-ui--kvm--hid-pak-secure: 1;

    /* Keyboard */
    --config-ui--kvm--hid-keyboard-bad-link: 0;
    --config-ui--kvm--hid-keyboard-swap-cc: 0;

    /* Mouse */
    --config-ui--kvm--hid-mouse-squash: 1;
    --config-ui--kvm--hid-mouse-reverse-scrolling: 0;
    --config-ui--kvm--hid-mouse-reverse-panning: 0;
    --config-ui--kvm--hid-mouse-cumulative-scrolling: 1;
    --config-ui--kvm--hid-mouse-dot: 1;
}
```

These set the **initial default** for new users or cleared browsers. Users can still override via the UI, and their preference is saved in localStorage as before.

## Notes

- Zero behavior change when no CSS custom properties are set — all original hardcoded values are preserved as fallbacks
- The `tools.config.getBool()` / `tools.config.get()` infrastructure and the CSS injection pipeline (`/etc/kvmd/web.css` → nginx `user.css` alias → `:root` properties) already exist
- This is particularly useful for kiosk deployments or multi-user setups where admins want consistent defaults

### DIFF
--- a/web/share/js/kvm/atx.js
+++ b/web/share/js/kvm/atx.js
@@ -39,7 +39,7 @@ export function Atx(__recorder) {
 		$("atx-power-led").title = "Power Led";
 		$("atx-hdd-led").title = "Disk Activity Led";
 
-		tools.storage.bindSimpleSwitch($("atx-ask-switch"), "atx.ask", true);
+		tools.storage.bindSimpleSwitch($("atx-ask-switch"), "atx.ask", tools.config.getBool("kvm--atx-ask", true));
 
 		tools.el.setOnClick($("atx-power-button"), () => __clickAtx("power"));
 		tools.el.setOnClick($("atx-power-button-long"), () => __clickAtx("power_long"));
--- a/web/share/js/kvm/hid.js
+++ b/web/share/js/kvm/hid.js
@@ -75,7 +75,7 @@ export function Hid(__getGeometry, __recorder) {
 			});
 		}
 
-		tools.storage.bindSimpleSwitch($("hid-sysrq-ask-switch"), "hid.sysrq.ask", true);
+		tools.storage.bindSimpleSwitch($("hid-sysrq-ask-switch"), "hid.sysrq.ask", tools.config.getBool("kvm--hid-sysrq-ask", true));
 
 		tools.el.setOnClick($("hid-jiggler-switch"), __clickJigglerSwitch);
 	};
--- a/web/share/js/kvm/keyboard.js
+++ b/web/share/js/kvm/keyboard.js
@@ -50,8 +50,8 @@ export function Keyboard(__recordWsEvent) {
 		window.addEventListener("focusin", __updateOnlineLeds);
 		window.addEventListener("focusout", __updateOnlineLeds);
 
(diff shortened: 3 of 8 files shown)


mdevaev:
Hello. I like the idea but I need to think a bit about implementation.

nullstacked:
> Hello. I like the idea but I need to think a bit about implementation.

Thank you for taking a look. Reason I went with this approach is because I saw you had code already using this already for one of the ui settings so I thought that was the established pattern:

`tools.storage.bindSimpleSwitch(
		$("page-full-tab-stream-switch"),
		"page.full_tab_stream",
		tools.config.getBool("kvm--full-tab-stream", false));`

Either way would be great to get a solution as the use case for this request is because I have multiple users across many browsers and we have to constantly reapply the settings.

mdevaev:
I plan to replace `storage` calls to some basic interface to avoid copying of config calls.

PR #212, opened 2026-05-06
feat: add support for vendor_id, product_id and product name for audio devices

rarspace01:
**Is your feature request related to a problem? Please describe.**
Currently, I can't configure the hardware ids for the audio devices.

**Describe the solution you'd like**
I would like the option in yaml to set the parameters accordingly to simulate real devices (and prevent Microsoft Entra Device endpoint problems).
**Describe alternatives you've considered**
Not using the audio

**Additional context**


mdevaev:
The options you are using do not exist. Where did you get this from?

rarspace01:
Sorry, my mistake. I should have double-checked in the kernel source and tested it locally.

PR #213, opened 2026-05-06
Add optional cipher suite parameter to IPMI GPIO pseudo-driver

gazzenger:
## Summary

With changes to the IPMI specification over the years, the default cipher suite utilised by the `ipmitool` library may not work with older servers that support earlier versions of the IPMI spec.

This PR extends the existing IPMI GPIO pseudo-driver to allow defining alternative cipher suites for connecting to a server. 

## What changed

Introduction of a new optional cipher suite ID parameter called `cipher` for the IPMI GPIO pseudo-driver for defining different cipher suites.

The cipher suite IDs are defined in Table 22-10 of the [IPMI v2 Specification](https://www.intel.la/content/dam/www/public/us/en/documents/specification-updates/ipmi-intelligent-platform-mgt-interface-spec-2nd-gen-v2-0-spec-update.pdf), ranging from ID 0 through 19, these IDs define the combination of algorithms used for authentication, integrity and confidentiality, when making the connection to a server.

Currently the `ipmitool` library defaults to using the cipher suite ID 17, hence this is the defined default in this PR. 

**Files changed:**
- `kvmd/plugins/ugpio/ipmi.py` — Psuedo-driver for IPMI

## Example usage

When configuring the GPIO config in PiKVM `override.yaml` file, the IPMI pseudo-driver can now be defined with the optional cipher suite.

```yaml
kvmd:
...
    gpio:
        drivers:
            my_server:
                type: ipmi
                host: myserver.local
                user: admin
                passwd: admin
                cipher: 3
...
```

## Notes

- This new parameter is optional, and is only required when interfacing with an IPMI interface that doesn't support the security ciphers.

### DIFF
--- a/kvmd/plugins/ugpio/ipmi.py
+++ b/kvmd/plugins/ugpio/ipmi.py
@@ -86,6 +86,7 @@ class Plugin(BaseUserGpioDriver):  # pylint: disable=too-many-instance-attribute
         return {
             "host":   Option("",  type=valid_ip_or_host),
             "port":   Option(623, type=valid_port),
+            "cipher": Option(17,  type=valid_number.mk(min=0, max=19)),
             "user":   Option(""),
             "passwd": Option(""),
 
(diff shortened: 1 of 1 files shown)


mdevaev:
Hello. Did you test this patch?

gazzenger:
Hi Maxim, 
I've tested this patch locally, and managed to find I was missing passing the `cipher` parameter through the constructor, so this has been fixed now.

gazzenger:
Looks like TOX CI is failing on an unrelated piece of code running during testing with mypy

mdevaev:
Don't worry, I'll fix it and accept your PR.

PR #219, opened 2026-05-31
Fix Python type hint syntax error in proto.py

hatch01:
## Problem
The application fails to start with a `SyntaxError` when Python tries to parse type hints using the `|` union operator with string literals in `kvmd/apps/kvmd/switch/proto.py` at line 133.

```
File "kvmd/apps/kvmd/switch/proto.py", line 133, in UnitState
  def compare_edid(self, ch: int, edid: ("Edid" | None)) -> bool:
                                        ~~~~~~~^~~~~~
SyntaxError: invalid syntax
```

## Root Cause
The file was missing `from __future__ import annotations`, which enables deferred evaluation of type hints. Without this import, Python attempts to evaluate `"Edid" | None` at runtime as actual code (using the bitwise OR operator), which is invalid syntax for a string and `None` type.

## Solution
Added `from __future__ import annotations` as the first import in the file (after the license header). This:
- Enables PEP 563 (deferred evaluation of annotations)
- Allows the modern `X | Y` union syntax (PEP 604) to work correctly
- Enables forward references without requiring string quotes

## Changes
- `kvmd/apps/kvmd/switch/proto.py`: Added `from __future__ import annotations` import

## Testing
- Verified the syntax error is resolved and the module can now be imported successfully

Co-authored with claude haiku 4.5

### DIFF
--- a/kvmd/apps/kvmd/switch/proto.py
+++ b/kvmd/apps/kvmd/switch/proto.py
@@ -19,6 +19,7 @@
 #                                                                            #
 # ========================================================================== #
 
+from __future__ import annotations
 
 import struct
 import dataclasses


mdevaev:
Hello. Which python version do you use?

hatch01:
I tested on 3.13 and 3.12 with the same behavior.
3.14 and 3.15 seems to have the same issue with this basic test : 
```
$ python3.15 -c '"Edid" | None'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    "Edid" | None
    ~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
$ python3.14 -c '"Edid" | None'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    "Edid" | None
    ~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

mdevaev:
Everything is fine on Python 3.14. The check you did is incorrect because `"Edid"` type wasn't defined at all. Moreover, it's not a type hint.

```
[root@pikvm ~]# cat test.py
class Foo:
    pass

def test(x: ("Foo" | None)):
    print(x)

test(Foo())

[root@pikvm ~]# python test.py
<__main__.Foo object at 0xffff85350c20>

[root@pikvm ~]# python --version
Python 3.14.5

[root@pikvm ~]# python3.14 -c '"Edid" | None'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    "Edid" | None
    ~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

hatch01:
Oh, that's completely my bad, I did not see that you updated to Python 3.14.

mdevaev:
No problem.

PR #222, opened 2026-06-07
Add randomization for mouse jiggler

spili-desu:
The mouse jiggler currently fires on a fixed interval (jiggler.interval, default 60s). A periodic jiggle is an easy signature for idle/anti-AFK detectors to fingerprint. This PR adds an optional randomizer that jitters the interval by a configurable +/- range, making the jiggle timing look more human and harder to detect.

Changes

- New config option kvmd.hid.jiggler.randomizer_interval (seconds, default 0 = disabled).
- When enabled, each jiggle waits a random duration in [interval − randomizer_interval, interval + randomizer_interval], floored at 1s, re-rolled after every fire.
- Added an info log line on each jiggle reporting the interval that was used.
- Exposed randomizer_interval in the jiggler state for introspection.

The configuration is YAML only, no UI change.

Documentation update in : https://github.com/pikvm/pikvm/pull/1666

### DIFF
--- a/kvmd/plugins/hid/__init__.py
+++ b/kvmd/plugins/hid/__init__.py
@@ -21,6 +21,7 @@
 
 
 import asyncio
+import random
 import time
 
 from typing import Final
@@ -31,10 +32,13 @@ from typing import Any
 
 from evdev import ecodes
 
+from ...logging import get_logger
+
 from ...yamlconf import Section
 from ...yamlconf import Option
 
 from ...validators.basic import valid_bool
+from ...validators.basic import valid_int_f0
 from ...validators.basic import valid_int_f1
 from ...validators.basic import valid_string_list
 from ...validators.hid import valid_hid_key
@@ -60,8 +64,10 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
 
         self.__j_enabled:  Final[bool]  = c.jiggler.enabled
         self.__j_interval: Final[float] = c.jiggler.interval
+        self.__j_randomizer_interval: Final[int] = c.jiggler.randomizer_interval
 
         self.__j_active: bool = c.jiggler.active
+        self.__j_next_interval: float = self.__roll_interval()
 
         self.__j_absolute = True
         self.__j_activity_ts = self.__get_monotonic_seconds()
@@ -84,6 +90,7 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
                 "enabled":  Option(True,  type=valid_bool),
                 "active":   Option(False, type=valid_bool),
                 "interval": Option(60,    type=valid_int_f1),
+                "randomizer_interval": Option(0, type=valid_int_f0),
             },
         }
 
@@ -251,6 +258,13 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
     def __get_monotonic_seconds(self) -> int:
         return int(time.monotonic())
 
+    def __roll_interval(self) -> float:
+        if self.__j_randomizer_interval <= 0:
+            return self.__j_interval
+        lo = max(1, int(self.__j_interval) - self.__j_randomizer_interval)
+        hi = int(self.__j_interval) + self.__j_randomizer_interval
+        return random.randint(lo, hi)
+
     def _set_jiggler_absolute(self, absolute: bool) -> None:
         self.__j_absolute = absolute
 
@@ -264,6 +278,7 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
                 "enabled":  self.__j_enabled,
                 "active":   self.__j_active,
                 "interval": self.__j_interval,
+                "randomizer_interval": self.__j_randomizer_interval,
             },
         }
 
@@ -271,7 +286,8 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
 
     async def systask(self) -> None:
         while True:
-            if self.__j_active and (self.__j_activity_ts + self.__j_interval < self.__get_monotonic_seconds()):
+            if self.__j_active and (self.__j_activity_ts + self.__j_next_interval < self.__get_monotonic_seconds()):
+                get_logger(0).info("Jiggling mouse (interval=%d seconds) ...", self.__j_next_interval)
                 if self.__j_absolute:
                     (x, y) = (self.__j_last_x, self.__j_last_y)
                     for move in (([100, -100] * 5) + [0]):
@@ -281,6 +297,7 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
                     for move in ([10, -10] * 5):
                         self.send_mouse_relative_event(move, move)
                         await asyncio.sleep(0.1)
+                self.__j_next_interval = self.__roll_interval()
             await asyncio.sleep(1)
 
 
--- a/kvmd/plugins/hid/__init__.py
+++ b/kvmd/plugins/hid/__init__.py
@@ -38,7 +38,6 @@ from ...yamlconf import Section
 from ...yamlconf import Option
 
 from ...validators.basic import valid_bool
-from ...validators.basic import valid_int_f0
 from ...validators.basic import valid_int_f1
 from ...validators.basic import valid_string_list
 from ...validators.hid import valid_hid_key
@@ -64,7 +63,6 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
 
         self.__j_enabled:  Final[bool]  = c.jiggler.enabled
         self.__j_interval: Final[float] = c.jiggler.interval
-        self.__j_randomizer_interval: Final[int] = c.jiggler.randomizer_interval
 
         self.__j_active: bool = c.jiggler.active
         self.__j_next_interval: float = self.__roll_interval()
@@ -90,7 +88,6 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
                 "enabled":  Option(True,  type=valid_bool),
                 "active":   Option(False, type=valid_bool),
                 "interval": Option(60,    type=valid_int_f1),
-                "randomizer_interval": Option(0, type=valid_int_f0),
             },
         }
 
@@ -259,10 +256,11 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
         return int(time.monotonic())
 
     def __roll_interval(self) -> float:
-        if self.__j_randomizer_interval <= 0:
-            return self.__j_interval
-        lo = max(1, int(self.__j_interval) - self.__j_randomizer_interval)
-        hi = int(self.__j_interval) + self.__j_randomizer_interval
+        # Randomize the wait by +/- 25% of the interval so the jiggling is less
+        # predictable (e.g. 60 -> +/-15, 120 -> +/-30, 30 -> +/-7).
+        jitter = int(self.__j_interval * 0.25)
+        lo = max(1, int(self.__j_interval) - jitter)
+        hi = int(self.__j_interval) + jitter
         return random.randint(lo, hi)
 
     def _set_jiggler_absolute(self, absolute: bool) -> None:
@@ -278,7 +276,6 @@ class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
                 "enabled":  self.__j_enabled,
                 "active":   self.__j_active,
                 "interval": self.__j_interval,
-                "randomizer_interval": self.__j_randomizer_interval,
             },
         }
 


mdevaev:
Hello. Before accepting this PR, I would ask you to take a look at this: https://github.com/Dapper-D/GhostHID/blob/main/src/ghost_sim.py

I would prefer to create some kind of universal algorithm and avoid additional configuration settings. I think the Bezier curve solution is good enough, and I was going to implement it. What do you think about this?

spili-desu:
Super interesting piece of code. Thank you for sharing @mdevaev 
Indeed, this implementation seems more robust and will obviously produce more "human" movements.

But for the temporal aspect, do you plan to do the same for interval too? I saw in ghost_sim.py:100 that the mouse moves every (30 to 120) seconds.

spili-desu:
Also, maybe the interval can be made a range to avoid adding another config keys?

mdevaev:
I think it makes sense to do some randomization here too.

There is another implementation using curves, but I don't remember where I found it. You can see my notes.
[j.py](https://github.com/user-attachments/files/28683976/j.py)

spili-desu:
So the new implementation would introduce a move_humanly function that produce random mouse actions (scroll, click, Bézier moves ... etc) ?

mdevaev:
Kinda.

spili-desu:
@mdevaev for this PR, i added another commit as a new proposition for the temporal randomization : simpler and without using a new config key. What do you think?

25% is arbitrary but seems to create good usable random ranges imho

mdevaev:
Looks fine for me. I can merge it as is, but if you want to play with curves - it's more than welcome.

spili-desu:
I would love to implement it - but if its okay for you - in another PR? I actually need this one for my professional usage quite soon. Thank you.

Edit : updated the doc PR too

mdevaev:
I don't mind if you do this in current PR.

mawadSur:
Hopping in as a fellow user of this feature 🙂 — I took a crack at the "universal, human-like" version you two discussed (@mdevaev's `ghost_sim.py` + the `j.py` notes), and while doing so I found why CI is red here. Two separate things below.

## 1. The failing CI check is a master-side bug, not this PR

The red **TOX CI** job dies while building the `testenv` Docker image, not in any of this PR's code:

```
#9 error: target not found: libjpeg
```

This PR only touches `kvmd/plugins/hid/__init__.py`, so it can't be the cause. What happened is upstream: Arch's `libjpeg-turbo` stopped satisfying the bare name `libjpeg` (it now only provides the versioned `libjpeg.so`), somewhere around 2026‑06‑28 → 07‑02. So `pacman -S libjpeg` in `testenv/Dockerfile` now fails. **`master`'s own CI is red with the identical error**, so a re-run won't help.

@mdevaev — the one-line fix is in `testenv/Dockerfile` (the big `$PACMAN -Syu …` list):

```diff
-		libjpeg \
+		libjpeg-turbo \
```

`libjpeg-turbo` is the real package that provides `libjpeg.so` (needed by ustreamer / pillow). It belongs as a small infra commit on `master` rather than in this feature branch — once it lands, this PR's CI should pass on re-run.

## 2. A "curves" implementation, if you want it

Since you said it'd be welcome in this PR, here's a version that keeps @spili-desu's already-approved ±25% interval jitter (no new config key) and swaps the fixed zig‑zag for a single **universal, human-like** gesture engine, taking the ideas from both references (random action picker from `ghost_sim.py` + cubic‑Bézier with a perpendicular arc & ease‑in/out from `j.py`):

- `__jiggle()` randomly picks **move / jitter / scroll** (weighted toward moves).
- `__jiggle_move()` glides to a nearby random point **and back** along two independent cubic‑Bézier arcs with ease‑in/out — implemented for **both absolute and relative** mouse modes.
- `__jiggle_jitter()` = a few micro‑twitches that settle back; `__jiggle_scroll()` = a net‑zero wheel nudge.
- Every gesture is **net‑zero displacement**, so the operator's real cursor never drifts. No new config keys — all internal, derived from the existing `interval`.

**Validation I ran** (no hardware needed for this part):
- Byte‑compiles on Python 3.14.
- 20k‑trial geometry check: max relative per‑step delta **22** (well under the ±127 HID limit, zero clamping), net drift **exactly 0**.
- Functional test importing the real methods: absolute moves return to base & stay within `MouseRange`; relative deltas ≤127 with zero net; scroll/jitter net‑zero; dispatcher runs cleanly in both modes; `interval` rolls within [45,75] for the 60 s default and never below 1 s.

Diff is **+117 / −12**, one file:

<details>
<summary>kvmd/plugins/hid/__init__.py — full diff</summary>

```diff
--- a/kvmd/plugins/hid/__init__.py
+++ b/kvmd/plugins/hid/__init__.py
@@ -21,6 +21,8 @@
 
 
 import asyncio
+import math
+import random
 import time
 
 from typing import Final
@@ -31,6 +33,8 @@
 
 from evdev import ecodes
 
+from ...logging import get_logger
+
 from ...yamlconf import Section
 from ...yamlconf import Option
 
@@ -43,12 +47,22 @@
 from ...keyboard.mappings import WEB_TO_EVDEV
 from ...keyboard.mappings import EvdevModifiers
 from ...mouse import MouseRange
+from ...mouse import MouseDelta
 
 from .. import BasePlugin
 from .. import get_plugin_class
 
 
 # =====
+# Jiggler motion tuning. These are internal constants (no config knobs) so the
+# jiggling stays a single "universal" algorithm that just looks human.
+_JIGGLE_STEPS:      Final[int]   = 40     # points traced along each Bezier segment
+_JIGGLE_STEP_DELAY: Final[float] = 0.02   # ~50 Hz, close to a real pointer's update rate
+_JIGGLE_MOVE_MIN:   Final[int]   = 50     # min nudge amplitude (HID units)
+_JIGGLE_MOVE_MAX:   Final[int]   = 400    # max nudge amplitude (HID units)
+
+
+# =====
 class BaseHid(BasePlugin):  # pylint: disable=too-many-instance-attributes
     def __init__(self, c: Section) -> None:
         super().__init__(c)
@@ -62,6 +76,7 @@
         self.__j_interval: Final[float] = c.jiggler.interval
 
         self.__j_active: bool = c.jiggler.active
+        self.__j_next_interval: float = self.__roll_interval()
 
         self.__j_absolute = True
         self.__j_activity_ts = self.__get_monotonic_seconds()
@@ -267,20 +282,110 @@
             },
         }
 
-    # =====
+    # ===== Jiggler =====
 
+    def __roll_interval(self) -> float:
+        # Jitter the wait by +/-25% of the configured interval so the jiggle
+        # timing is not a fixed period an idle/anti-AFK detector can fingerprint
+        # (e.g. 60 -> [45, 75], 120 -> [90, 150], 30 -> [23, 37]). Re-rolled
+        # after every jiggle. No extra config knob -- derived from `interval`.
+        jitter = self.__j_interval * 0.25
+        return max(1.0, random.uniform(self.__j_interval - jitter, self.__j_interval + jitter))
+
+    async def __jiggle(self, absolute: bool) -> None:
+        # A single "universal" human-like anti-idle action. Instead of a fixed
+        # zig-zag we pick one of a few natural gestures at random; the pointer
+        # always returns to where it started, so the operator's cursor doesn't
+        # drift over time.
+        action = random.choice(("move", "move", "move", "jitter", "scroll"))
+        if action == "move":
+            await self.__jiggle_move(absolute)
+        elif action == "jitter":
+            await self.__jiggle_jitter(absolute)
+        else:
+            await self.__jiggle_scroll()
+
+    async def __jiggle_move(self, absolute: bool) -> None:
+        # Glide to a nearby random point and back along curved (cubic Bezier)
+        # paths with ease-in/out timing -- the two legs use independent random
+        # arcs, so it looks like a hand nudging the mouse rather than a machine.
+        amplitude = random.randint(_JIGGLE_MOVE_MIN, _JIGGLE_MOVE_MAX)
+        angle = random.uniform(0.0, math.tau)
+        dest = (round(amplitude * math.cos(angle)), round(amplitude * math.sin(angle)))
+        path = self.__bezier((0, 0), dest) + self.__bezier(dest, (0, 0))
+        if absolute:
+            (base_x, base_y) = (self.__j_last_x, self.__j_last_y)
+            for (off_x, off_y) in path:
+                self.send_mouse_move_event(
+                    MouseRange.normalize(base_x + off_x),
+                    MouseRange.normalize(base_y + off_y),
+                )
+                await asyncio.sleep(_JIGGLE_STEP_DELAY)
+        else:
+            (prev_x, prev_y) = (0, 0)
+            for (off_x, off_y) in path:
+                self.send_mouse_relative_event(
+                    MouseDelta.normalize(off_x - prev_x),
+                    MouseDelta.normalize(off_y - prev_y),
+                )
+                (prev_x, prev_y) = (off_x, off_y)
+                await asyncio.sleep(_JIGGLE_STEP_DELAY)
+
+    async def __jiggle_jitter(self, absolute: bool) -> None:
+        # A handful of tiny twitches in place, then settle back to the start.
+        (base_x, base_y) = (self.__j_last_x, self.__j_last_y)
+        for _ in range(random.randint(3, 6)):
+            (off_x, off_y) = (random.randint(-2, 2), random.randint(-2, 2))
+            if absolute:
+                self.send_mouse_move_event(
+                    MouseRange.normalize(base_x + off_x),
+                    MouseRange.normalize(base_y + off_y),
+                )
+            else:
+                self.send_mouse_relative_event(off_x, off_y)
+            await asyncio.sleep(random.uniform(0.05, 0.15))
+        if absolute:
+            self.send_mouse_move_event(base_x, base_y)  # Settle exactly back
+
+    async def __jiggle_scroll(self) -> None:
+        # A small wheel nudge one way and back (net-zero).
+        direction = random.choice((-1, 1))
+        self.send_mouse_wheel_event(0, direction)
+        await asyncio.sleep(random.uniform(0.1, 0.3))
+        self.send_mouse_wheel_event(0, -direction)
+
+    def __bezier(self, start: tuple[int, int], end: tuple[int, int]) -> list[tuple[int, int]]:
+        # Cubic Bezier from start to end with a random perpendicular "bow" and
+        # ease-in/out timing, returned as integer points. Control points sit at
+        # 25% and 75% along the straight line, pushed sideways by a random
+        # fraction of the distance so the trajectory curves like a real hand.
+        (sx, sy) = start
+        (ex, ey) = end
+        (vx, vy) = (ex - sx, ey - sy)
+        dist = math.hypot(vx, vy)
+        if dist < 1:
+            return [end]
+        (perp_x, perp_y) = (-vy / dist, vx / dist)
+        bow = random.uniform(-0.3, 0.3) * dist
+        c1 = (sx + vx * 0.25 + perp_x * bow, sy + vy * 0.25 + perp_y * bow)
+        c2 = (sx + vx * 0.75 + perp_x * bow, sy + vy * 0.75 + perp_y * bow)
+        points: list[tuple[int, int]] = []
+        for i in range(1, _JIGGLE_STEPS + 1):
+            t = i / _JIGGLE_STEPS
+            # EaseInOutQuad: accelerate to the midpoint, then decelerate.
+            e = (2 * t * t) if t < 0.5 else (-1 + (4 - 2 * t) * t)
+            mt = 1 - e
+            x = mt ** 3 * sx + 3 * mt ** 2 * e * c1[0] + 3 * mt * e ** 2 * c2[0] + e ** 3 * ex
+            y = mt ** 3 * sy + 3 * mt ** 2 * e * c1[1] + 3 * mt * e ** 2 * c2[1] + e ** 3 * ey
+            points.append((round(x), round(y)))
+        return points
+
     async def systask(self) -> None:
         while True:
-            if self.__j_active and (self.__j_activity_ts + self.__j_interval < self.__get_monotonic_seconds()):
-                if self.__j_absolute:
-                    (x, y) = (self.__j_last_x, self.__j_last_y)
-                    for move in (([100, -100] * 5) + [0]):
-                        self.send_mouse_move_event(MouseRange.normalize(x + move), MouseRange.normalize(y + move))
-                        await asyncio.sleep(0.1)
-                else:
-                    for move in ([10, -10] * 5):
-                        self.send_mouse_relative_event(move, move)
-                        await asyncio.sleep(0.1)
+            if self.__j_active and (self.__j_activity_ts + self.__j_next_interval < self.__get_monotonic_seconds()):
+                get_logger(0).info("Jiggling mouse (interval=%.1f seconds) ...", self.__j_next_interval)
+                await self.__jiggle(self.__j_absolute)
+                self.__j_next_interval = self.__roll_interval()
             await asyncio.sleep(1)
 
 
```
</details>

Happy to shape it differently, trim the action variety, or split it into its own PR if you'd rather keep this one as the minimal interval‑randomizer. (Also minor: the PR description still documents the removed `randomizer_interval` config option — worth updating to the ±25% approach.)

<sub>Reviewed/prototyped with Claude Code; geometry & behavior validated offline, not yet on real HID hardware.</sub>

mawadSur:
### Follow-up: tested on real PiKVM hardware ✅

I ran the curves implementation on an actual **PiKVM V4 (CM4)** — kvmd `4.171`, Python `3.14.5` — to confirm it behaves on the real HID runtime, not just on paper.

Method: dropped the patched `BaseHid` in as a throwaway module inside the installed `kvmd` package (so its relative imports and the **real** `kvmd.mouse.MouseRange`/`MouseDelta` + `evdev` are exercised), then drove the actual `__jiggle*` methods and captured the exact mouse-event stream each gesture emits. The live `hid/__init__.py` was never modified and the module was removed afterward.

**Results — all green:**

```
device python: 3.14.5   kvmd: 4.171   MouseRange (-32768,32767)   MouseDelta (-127,127)

sample ABSOLUTE move: 80 events, first 4 [(0,0),(0,-1),(-1,-2),(-1,-4)] ... last (0,0)
sample RELATIVE move: 80 events, max |delta| = 20, net = (0,0)

  [PASS] byte-compiles + imports inside the real kvmd package
  [PASS] absolute move returns exactly to base (0,0)
  [PASS] absolute points within MouseRange
  [PASS] relative deltas within +/-127 (never truncated on the wire)
  [PASS] relative net displacement zero (no cursor drift)
  [PASS] scroll is a net-zero up/down pair
  [PASS] absolute jitter settles back to base
  [PASS] full __jiggle dispatch runs cleanly (both modes)
  [PASS] interval rolls within +/-25% band [45,75] for 60s
```

Since every gesture goes through the same `send_mouse_move_event` / `send_mouse_relative_event` / `send_mouse_wheel_event` entry points the existing jiggler already uses (with values inside the same ranges), it's drop-in compatible with the HID backends. A sample absolute round-trip path — out along one arc, back along a different one (`S` = start/return, `X` = far point):

```
amplitude=282  dest=(267,-91)  points=80
+------------------------------------------------+
|                                        ......X |
|                                  . .. .     .. |
|                                .            .  |
|                           .  .             .   |
|                         .                  .   |
|                    . .                    .    |
|                 .                        .     |
|               .                         .      |
|           . .                          .       |
|         .                             .        |
|      . .                            .          |
|    ..                             .            |
|  ..                             .              |
|S.                             .                |
|...                          .                  |
|  . ..                  . .                     |
|      .. .. . . .  . .                          |
+------------------------------------------------+
```

I deliberately did **not** activate the live jiggler against a connected target (it would move that machine's real cursor). Happy to do a brief live run on request, or to shape the gesture mix / amplitudes however you'd prefer.

mdevaev:
@spili-desu I'll merge it as-is. Maybe I'll make the curves myself, or you can come back to this question if you're interested again.

@mawadSur thank you too for the draft implementation of curves.

v1ckxy:
After latest update, this pr broke jiggler for me 🤨, it was jiggling basically every time I wasn't moving the mouse 🤣

So I changed `interval = 0.0` to `interval = self.__j_interval` in this line:
https://github.com/pikvm/kvmd/blob/master/kvmd/plugins/hid/__init__.py#L276

And it looks like it works ok now.
Not sure if I botched something but at least it waits at least `__j_interval` seconds before doing anything.

mdevaev:
@v1ckxy fixed and released, thank you.

PR #225, opened 2026-06-19
pstrun: restore terminal foreground process group after the child exits

### LINKED ISSUE #1467: kvmd-certbot doesn't accept interactive user input because the 2nd invocation of kvmd-pstrun fails with tcsetpgrp()
**Describe the bug**
kvmd-certbot runs kvmd-pstrun twice: [one](https://github.com/pikvm/kvmd/blob/beb5d541b03be044f03125578d5a1b77064c197b/scripts/kvmd-certbot#L52), [two](https://github.com/pikvm/kvmd/blob/beb5d541b03be044f03125578d5a1b77064c197b/scripts/kvmd-certbot#L99). kvmd-pstrun creates a process group and sets it as foreground for the controlling terminal: [source](https://github.com/pikvm/kvmd/blob/beb5d541b03be044f03125578d5a1b77064c197b/kvmd/apps/pstrun/__init__.py#L45-L48). Something goes wrong on the second invocation of kvmd-pstrun, and certbot doesn't see any user input when it asks interactive questions.

Commenting out `ensure_runroot` or changing it to `ensure_runroot < /dev/null > /dev/null 2>&1` works around the problem by skipping `tcsetpgrp()` in the first invocation of kvmd-pstrun, and leaves the terminal functional to `kvmd-pstrun -- certbot certonly ...`.

**To Reproduce**
1. ssh into the pikvm interactively (with pty allocated, i.e. `ssh -t`, which is the default when the command is not passed immediately)
2. Run `kvmd-certbot certonly_webroot --email 'your@email.com' --domain 'your.domain.com'`. If you don't have a certificate deployed, it should ask some questions, such as whether you agree to the terms of service and whether you'd like to subscribe to EFF. In my case, I already have a certificate, and it doesn't expire soon, so it asks me whether I'd like to replace it or keep the old one. It doesn't matter, as long as it asks you some question.
3. Try to provide an answer to certbot. Observe that it doesn't react and doesn't see anything in stdin (strace shows it stays in pselect6 waiting on fd 0, and /proc/.../fd/0 points to /dev/pts/1).

**Expected behavior**
The process that runs in kvmd-pstrun should be able to accept user input from stdin, even if it's the second invocation of kvmd-pstrun in a bash script.

**Screenshots**
```
# kvmd-certbot certonly_webroot --email '**********' --domain '**********' --agree-tos --no-eff-email
--    INFO -- Opening PST session ...
--    INFO -- PST write is allowed: /var/lib/kvmd/pst
--    INFO -- Running the process ...
+ mkdir -p /var/lib/kvmd/pst/data/certbot/runroot
+ chown -R kvmd-certbot: /var/lib/kvmd/pst/data/certbot/runroot/..
--    INFO -- Process finished: returncode=0
--    INFO -- Opening PST session ...
--    INFO -- PST write is allowed: /var/lib/kvmd/pst
--    INFO -- Running the process ...
Saving debug log to /var/lib/kvmd/pst/data/certbot/runroot/logs/letsencrypt.log
Certificate not yet due for renewal

You have an existing certificate that has exactly the same domains or certificate name you requested and isn't close to expiry.
(ref: /var/lib/kvmd/pst/data/certbot/runroot/config/renewal/**********.conf)

What would you like to do?
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
1: Keep the existing certificate for now
2: Renew & replace the certificate (may be subject to CA rate limits)
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Select the appropriate number [1-2] then [enter] (press 'c' to cancel): c
c
c
```
No reaction until I `killall certbot` from another terminal, then:
```
--    INFO -- Process finished: returncode=-15
# c
-bash: c: command not found
# c
-bash: c: command not found
# c
-bash: c: command not found
```

**Desktop (please complete the following information):**
 - OS: Arch Linux

**PiKVM info:**
  - Raspberry Pi board version: RPi 4
  - PiKVM platform: v3 and v4 plus
  - Video capture type: not relevant
  - KVMD version:
    ```
    kvmd 4.52-1
    kvmd-fan 0.32-1
    kvmd-platform-v4plus-hdmi-rpi4 4.52-1
    kvmd-webterm 0.50-1
    ```
  - uStreamer version: `ustreamer 6.29-1`
  - Linux kernel: `6.6.45-10-rpi #1 SMP Fri Jan 24 03:51:01 UTC 2025 armv7l GNU/Linux`

**Additional context**

mawadSur:
## Problem

Fixes pikvm/pikvm#1467.

`kvmd-pstrun` runs its child in its own **foreground process group** so interactive programs (like `certbot`) can talk to the terminal — `_preexec()` does `os.setpgrp()` followed by `os.tcsetpgrp(0, …)`, mimicking what a job-control shell does for a foreground job:

https://github.com/pikvm/kvmd/blob/master/kvmd/apps/pstrun/__init__.py#L43-L49

But unlike a shell, `kvmd-pstrun` **never hands the terminal back** when the child exits. After the child is gone, the controlling terminal's foreground process group still points at the now-dead child group, so any process that reads stdin afterwards is no longer in the foreground group and never receives user input.

`kvmd-certbot` invokes `kvmd-pstrun` twice — first `ensure_runroot`, then the actual `certbot` run:

https://github.com/pikvm/kvmd/blob/master/scripts/kvmd-certbot#L51-L57
https://github.com/pikvm/kvmd/blob/master/scripts/kvmd-certbot#L99

So the first invocation leaves the terminal in a broken state, and the second invocation's interactive `certbot` prompts (ToS agreement, "keep/renew existing certificate?", etc.) never see the user's keystrokes. The reporter (@gentoo-root) confirmed that forcing the first invocation to skip `tcsetpgrp()` (via `ensure_runroot < /dev/null`) works around it.

## Fix

Capture the terminal's foreground process group just before launching the child, and restore it once the child has finished — exactly the step a job-control shell performs after a foreground job exits. `SIGTTOU` is ignored around the restoring `tcsetpgrp()` (the parent is a background group at that point), consistent with the existing handling in `_run_process()`.

This neutralizes the lingering terminal side effect that the reporter's `< /dev/null` workaround avoids, and also leaves the terminal clean for the `--` passthrough and `renew` code paths, not just `certonly`.

## Notes / testing

The change is minimal and mirrors the existing style in the file. I was not able to run a full end-to-end test of the PiKVM + PST-daemon + `sudo`-to-`kvmd-certbot` + interactive `ssh -t` stack myself; verification here is by static reasoning + syntax compile. Happy to adjust placement (e.g. capture/restore inside `_run_process` instead) if you'd prefer.

### DIFF
--- a/kvmd/apps/pstrun/__init__.py
+++ b/kvmd/apps/pstrun/__init__.py
@@ -49,6 +49,29 @@ def _preexec() -> None:
             get_logger(0).info("Can't perform tcsetpgrp(0): %s", tools.efmt(ex))
 
 
+def _get_fg_pgrp() -> int:
+    if os.isatty(0):
+        try:
+            return os.tcgetpgrp(0)
+        except Exception as ex:
+            get_logger(0).info("Can't perform tcgetpgrp(0): %s", tools.efmt(ex))
+    return -1
+
+
+def _restore_fg_pgrp(pgrp: int) -> None:
+    # _preexec() puts the child into its own foreground process group via tcsetpgrp().
+    # When the child exits, the controlling terminal still points at that now-dead group,
+    # so any subsequent process that reads stdin (e.g. the second kvmd-pstrun launched by
+    # kvmd-certbot) never receives the user's input. Hand the terminal back to the original
+    # foreground group, exactly like a job-control shell does after a foreground job exits.
+    if pgrp >= 0 and os.isatty(0):
+        signal.signal(signal.SIGTTOU, signal.SIG_IGN)
+        try:
+            os.tcsetpgrp(0, pgrp)
+        except Exception as ex:
+            get_logger(0).info("Can't restore tcsetpgrp(0): %s", tools.efmt(ex))
+
+
 async def _run_process(cmd: list[str], data_path: str) -> asyncio.subprocess.Process:  # pylint: disable=no-member
     # https://stackoverflow.com/questions/58918188/why-is-stdin-not-propagated-to-child-process-of-different-process-group
     if os.isatty(0):
@@ -68,6 +91,7 @@ async def _run_cmd_ws(cmd: list[str], ws: aiohttp.ClientWebSocketResponse) -> in
     recv_task: (asyncio.Task | None) = None
     proc_task: (asyncio.Task | None) = None
     proc: (asyncio.subprocess.Process | None) = None  # pylint: disable=no-member
+    fg_pgrp = -1
 
     try:  # pylint: disable=too-many-nested-blocks
         while True:
@@ -87,6 +111,7 @@ async def _run_cmd_ws(cmd: list[str], ws: aiohttp.ClientWebSocketResponse) -> in
                         if event["data"]["write_allowed"] and proc is None:
                             logger.info("PST write is allowed: %s", event["data"]["path"])
                             logger.info("Running the process ...")
+                            fg_pgrp = _get_fg_pgrp()
                             proc = await _run_process(cmd, event["data"]["path"])
                         elif not event["data"]["write_allowed"]:
                             logger.error("PST write is not allowed")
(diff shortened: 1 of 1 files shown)


mdevaev:
Why you closed this? I was going to merge it.

mawadSur:
Was doing some clean up and accidently deleted the fork that I had, opening it back up. (This is why coding at 1am isn't  good idea) Should be up shortly.

mawadSur:
Had to open a new one becuase I deleted the repo, sorry
https://github.com/pikvm/kvmd/pull/228

PR #227, opened 2026-07-03
Add drawing tablet/pen input method.

chuck-r:
This commit adds the capability to use a graphics tablet pen as a mouse input in PiKVM. Since a graphics tablet can hover over the surface to move the cursor, tapping the tablet can be seen as a click. This is in contrast to the ambiguity and imprecision of touch input. Right clicks are achieved either by a pen button, or by holding the stylus in-place (within the drag threshold area) for a configurable period of time (right-click delay) before lifting. Left-click-and-drag is achieved by touching and holding the screen, followed by moving the cursor outside of the configurable area (drag threshold).

Since the DOM pointer events are also fired for mouse, touch, etc., I made sure to make these new event handlers handle pen input exclusively, so as not to create unintended consequences for either touch or mouse inputs. Touch and mouse inputs are unchanged.

**Possible improvements**
* UI placement: I wasn't really sure where the best spot for these options was so I packed them under the interface settings, but there may be a more reasonable place for them.
* ~~Near-the-cursor indicator of the right-click status as a visual cue indicating whether releasing will produce a left or a right click action at the moment. Windows Apps (the Microsoft RDP client) uses a hollow circle indicator that progresses from no icon, to a quarter, half, etc. until full to indicate that release of the stylus will produce a left click (incomplete circle) or right click (full circle). I'm not graphic designer, so this might best be left for another contributor.~~ ✅**Done**
* ~~Pen input only makes sense as an option in absolute cursor mode, but I couldn't find a way to cleanly enable/disable pen input depending on the HID mode. The HID radio input isn't populated until after the __init__ phase of mouse.js, so I couldn't find an easy way to hook the onChange event of hid-mouse-output-radio within my code to update the pen input availability dynamically. So, my code ignores pen input if __abs is false within mouse.js, but there is no obvious indication to the user.~~ ✅**Done**
* I made the drag area a square for easy computation, but could make it a circle instead for more diagonal consistency.

Tested with both a Wacom Intuos tablet and a Galaxy Tab S10 FE.
* The Galaxy Tab S10 FE stylus does not have a button that can be captured by the browser, so it effectively doesn't have a button. This is the reason behind the right-click emulation.
* The Wacom Intuos stylus' first button acts as a right click by default (which makes sense to me) and doesn't require right-click emulation: hence the option to disable right-click emulation.


mdevaev:
You don't need to write html manually, please make it in the pug files and run `make pug` to generate index.html from them.

chuck-r:
Sorry about the rapid commits. I think it's done now =)

mdevaev:
No problem)

PS: Could you take a look at https://github.com/pikvm/pikvm/issues/581?

chuck-r:
Sorry, I haven't had a chance to examine it yet. I'm not in that industry any more. I do have a friend that sees it often, but I can't afford to give him my current PiKVM for an indeterminate period of time. If I could find my old V3, I'd happily lend it to him, but alas it has gone missing.

mdevaev:
It's sad, because it seems it should be fixed now. I could test it myself, but I don't have any friends with xbox :)

chuck-r:
Also, wanted to let you know that `make pug` didn't work for me. It gave me an error when installing libjpeg. On my Manjaro (Arch-based) system, only libjpeg-turbo seems to be available, so I modified the docker file to use libjpeg-turbo instead of libjpeg and it seems to have worked.

I did find this reference to deprecation of the libjpeg virtual package: https://archlinux.org/todo/stop-relying-on-virtual-libjpeg-lib32-libjpeg-package/

mdevaev:
Make git rebase from upstream, I fixed it some time ago.

PR #230, opened 2026-07-03
ipmi: fix SOL console stutter by decoupling serial reads from the blocking sender (#899)

### LINKED ISSUE #899: ipmi sol support stutters
**Describe the bug**
Using ipmi sol support, the serial output stutters.  For instance when viewing the console of a Linux system during boot or reviewing boot logs with dmesg, I'll get anywhere from a few lines to a few screen fulls of text before a brief pause and the output continues.  This is disruptive to use as a serial console server, which should be able to smoothly stream the console log, and as found on other hardware BMC devices or when directly using a terminal program such as minicom.

**To Reproduce**
Steps to reproduce the behavior, like:
1. Configure IPMI SOL support as described in https://docs.pikvm.org/ipmi/#ipmi-bmc
2. Use a USB serial adapter, such as https://www.amazon.com/dp/B075YHFMC7
3. Note, add kvmd-ipmi to the uucp group to access USB serial (ex. gpasswd -a kvmd-ipmi uucp)
4. Configure all of the above at 115200 baud rate and connect to Linux host system configured for the same.
5. Use ipmitool on the pikvm system to connect to SOL: ex. ipmitool -I lanplus -U admin -P admin -H localhost sol activate
6. Run 'dmesg' from the SOL session

**Expected behavior**
Text should stream smoothly.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Fedora Linux
 - Browser: N/A
 - Version:N/A
 - VNC client: N/A

**PiKVM info:**
  - Raspberry Pi board version [e.g. RPi 4]: RPI4
  - PiKVM platform [e.g. v2-hdmi]: v2-hdmiusb
  - Video capture type [e.g. CSI bridge]: USB
  - KVMD version: `pacman -Q | grep kvmd`: kvmd 3.173-1, kvmd-fan 0.22-1, kvmd-oled 0.13-1, kvmd-platform-v2-hdmiusb-rpi4 3.173-1, kvmd-webterm 0.42-1
  - uStreamer version: `pacman -Q | grep ustreamer`: ustreamer 5.31-1
  - Linux kernel: `uname -a`: Linux pikvm 5.15.68-3-rpi-ARCH #1 SMP Mon Oct 31 20:56:54 MSK 2022 armv7l GNU/Linux

**Additional context**
Maxim kindly logged into my pikvm system to see the issue first hand and asked me to file this issue.  Thanks!

mawadSur:
## Problem

Fixes pikvm/pikvm#899.

IPMI SOL console output through `kvmd-ipmi` stutters: a few lines to a few screenfuls
appear, then a brief pause, repeatedly (e.g. running `dmesg` over `ipmitool ... sol activate`
at 115200 baud). It's unusable as a smooth serial console compared to a direct terminal or
`conserver`.

## Root cause

kvmd's SOL worker (`kvmd/apps/ipmi/server.py`) ran a single-threaded loop:

```python
while not stop:
    ready = select.select([qr, tty], ...)
    if tty in ready:
        self.__sol_console.send_data(tty.read_all())   # <-- blocks here
```

`self.__sol_console` is a `pyghmi.ipmi.console.ServerConsole`, whose `send_data()` inherits
`Console`'s **stop-and-wait** SOL sender (`console.py` `_sendoutput`): it transmits a single
packet (≤ `maxoutcount`, hard-coded to **256 bytes**) and then **blocks the calling thread**
in a retry loop (`while retries and self.awaitingack: self.wait_for_rsp(0.5); ...`) until the
client ACKs — only then returning to send the next 256-byte chunk.

Because that blocking send runs **inline in the read loop**, the serial port is not read for
the whole ACK round-trip. Serial data piles up in the kernel buffer and floods out on the next
`read_all()`, and user keystrokes aren't serviced either — i.e. bursts-then-pause + input lag.

## Fix

Decouple reading from sending. A dedicated **sender thread** owns every (blocking)
`send_data()` call, while the worker thread keeps draining the serial port and forwarding
user input continuously. Only the sender thread ever touches the pyghmi console, so the
session pump is still driven by exactly one extra thread (as before) and the console's output
lock is never contended. (Refactored into a module-level `_sol_pump()` so it's unit-testable.)

Also exposes **`ipmi.sol.max_packet`** (default `256`, i.e. unchanged) to raise pyghmi's
per-packet SOL cap. Each packet costs one blocking ACK round-trip, so a larger cap means fewer
round-trips for the same data. It's left at pyghmi's default because some SOL clients may
reject oversized payloads — operators who control their client can raise it and test.

### Honest scope

The **window-of-1** behavior is mandated by the IPMI 2.0 SOL spec, and the 256-byte cap and the
0.5s retransmit timer live inside pyghmi. So this change removes the read-starvation/burstiness
and input lag and gives operators the packet-size lever, but it does **not** remove the
fundamental round-trip-bound throughput ceiling — SOL gets noticeably smoother/more responsive
but won't match a raw TCP stream like `conserver`. A complete cure (adaptive retransmit,
negotiated packet size, non-blocking send) belongs upstream in pyghmi. Happy to adjust the
approach if you'd prefer it shaped differently.

## Tests

Adds `testenv/tests/apps/ipmi/test_sol.py`. It drives the real `_sol_pump()` against a
`os.openpty()` serial device with a stub console whose `send_data()` **blocks** (modelling
pyghmi's ACK wait), and asserts the reader keeps draining during the blocked send (max gap
between serial reads ≪ the block time) with no data loss.

I could not run the Docker `testenv` (`make tox`) on my machine (macOS — it needs Linux
`evdev`/`libxkbcommon`), so I validated `_sol_pump()` in isolation against pyghmi 1.6.18 +
pyserial 3.5 on a pty. Before/after, feeding a pty at ~11520 B/s for 3s with `send_data()`
blocking 1.0s per call:

| loop | fed | read | reads | **max read gap** |
|------|-----|------|-------|------------------|
| old (inline send) | 3168 B | 3168 B | 5 | **1005 ms** (starved; back-pressures the source to ~1 KB/s) |
| new (`_sol_pump`) | 34560 B | 34560 B | 360 | **10 ms** (full rate drained during blocked sends) |

What this does **not** cover (needs hardware I don't have): the real over-the-wire RTT/jitter,
the specific USB-serial adapter timing, and whether a raised `max_packet` is tolerated by a
given client. Please run CI and, if possible, a real `ipmitool` SOL session to confirm.

### DIFF
--- a/kvmd/apps/_scheme.py
+++ b/kvmd/apps/_scheme.py
@@ -704,6 +704,7 @@ def make_config_scheme() -> dict:
                 "speed":          Option(115200, type=valid_tty_speed, unpack_as="sol_speed"),
                 "select_timeout": Option(0.1,    type=valid_float_f01, unpack_as="sol_select_timeout"),
                 "proxy_port":     Option(0,      type=valid_port, unpack_as="sol_proxy_port"),
+                "max_packet":     Option(256,    type=valid_int_f1, unpack_as="sol_max_packet"),
             },
         },
 
--- a/kvmd/apps/ipmi/server.py
+++ b/kvmd/apps/ipmi/server.py
@@ -27,6 +27,8 @@ import threading
 import functools
 import queue
 
+from typing import Callable
+
 import aiohttp
 import serial
 
@@ -43,6 +45,68 @@ from ... import aiomulti
 from ... import network
 
 from .auth import IpmiAuthManager
+
+
+# =====
+def _sol_pump(
+    console: IpmiConsole,
+    tty: serial.Serial,
+    user_q: aiomulti.AioMpQueue[bytes],
+    select_timeout: float,
+    should_stop: Callable[[], bool],
+) -> None:
+
+    # pyghmi's ServerConsole.send_data() is a blocking, stop-and-wait SOL sender: it
+    # transmits a single (<=maxoutcount bytes) SOL packet and then waits for the client's
+    # ACK before returning. Calling it inline in the read loop stalls serial reads for the
+    # whole ACK round-trip, so the kernel serial buffer fills up and console output arrives
+    # in bursts separated by pauses while keystrokes lag (pikvm/pikvm#899).
+    #
+    # Decouple reading from sending: a dedicated sender thread owns every (blocking)
+    # send_data() call, while this thread keeps draining the serial port and forwarding
+    # user input without interruption. Only the sender thread ever touches the console, so
+    # pyghmi's session pump is still driven by exactly one extra thread (as before) and its
+    # output lock is never contended.
+
+    out_buf = bytearray()
+    out_lock = threading.Lock()
+    out_wake = threading.Event()
+
+    def sender() -> None:
+        while not should_stop():
+            out_wake.wait(select_timeout)
+            out_wake.clear()
+            with out_lock:
+                if not out_buf:
+                    continue
+                chunk = bytes(out_buf)
+                out_buf.clear()
+            console.send_data(chunk)  # Blocks until the client ACKs; only this thread sends
+
+    sender_thread = threading.Thread(target=sender, daemon=True)
+    sender_thread.start()
+    try:
+        qr = user_q.get_reader()
+        while not should_stop():
+            ready = select.select([qr, tty], [], [], select_timeout)[0]
+            if qr in ready:
+                data = b""
+                for _ in range(user_q.qsize()):  # Don't hold on this with [not empty()]
+                    try:
+                        data += user_q.get_nowait()
+                    except queue.Empty:
+                        break
+                if data:
+                    tty.write(data)
+            if tty in ready:
+                data = tty.read_all()
+                if data:
+                    with out_lock:
+                        out_buf += data
+                    out_wake.set()  # Hand the new data to the sender
+    finally:
+        out_wake.set()  # Wake the sender so it observes should_stop() and exits
+        sender_thread.join()
 
 
 # =====
@@ -63,6 +127,7 @@ class IpmiServer(BaseIpmiServer):  # pylint: disable=too-many-instance-attribute
         sol_speed: int,
         sol_select_timeout: float,
         sol_proxy_port: int,
+        sol_max_packet: int,
     ) -> None:
 
         host = network.get_listen_host(host)
@@ -79,6 +144,7 @@ class IpmiServer(BaseIpmiServer):  # pylint: disable=too-many-instance-attribute
         self.__sol_speed = sol_speed
         self.__sol_select_timeout = sol_select_timeout
         self.__sol_proxy_port = (sol_proxy_port or port)
+        self.__sol_max_packet = sol_max_packet
 
         self.__sol_lock = threading.Lock()
         self.__sol_console: (IpmiConsole | None) = None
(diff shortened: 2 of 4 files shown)


mawadSur:
## Tested on real PiKVM hardware ✅

I got hold of a PiKVM and validated this on-device (the original PR was written without hardware access). Short version: the core fix and the new `_sol_pump` regression test pass on the real target runtime, an end-to-end `ipmitool` SOL session works, and I was able to pin down exactly how far `ipmi.sol.max_packet` can be pushed before a real client breaks.

**Device / runtime**
- PiKVM V4 (Raspberry Pi Compute Module 4), Arch Linux ARM, kvmd **4.171**
- Python **3.14.5**, pyserial **3.5**, pyghmi **1.6.11**, ipmitool **1.8.19**
- Patch applied cleanly: `apps/ipmi/server.py` is byte-identical to this PR's base, `_scheme.py` took the one-line `max_packet` addition, `valid_int_f1` was already imported.

### 1. New regression test passes on-device
Ran `testenv/tests/apps/ipmi/test_sol.py`'s `_sol_pump` logic against the device's real Python/pyserial, importing the **installed patched** `_sol_pump`:
- Committed scenario (send blocks 0.5 s): max read-gap **8.9 ms** (threshold 250 ms), **0 bytes lost** — all 5 assertions PASS.
- Before/after contrast (send blocks 1.0 s, 3 s feed at 115200 baud):

| loop | fed | read | reads | sends | max read gap |
|---|---:|---:|---:|---:|---:|
| old (inline send) | 27936 | 20571 | 6 | 6 | **1000 ms** |
| new (`_sol_pump`) | 34560 | 34560 | 360 | 4 | **9 ms** |

The old inline path stalls a full second per send and back-pressures the serial source; the new pump keeps draining throughout the blocked send.

### 2. Config + integration
`ipmi.sol.max_packet` validates on-device (default `256`, unpacks to `sol_max_packet`, rejects `0`/`-5`/non-int), flows into `IpmiServer.__init__`, and `kvmd-ipmi` starts and reaches *"Listening IPMI …"* with the patched code. Reading pyghmi 1.6.11: `ServerConsole.maxoutcount` (default 256) is the outbound chunk cap (`console.py:251`) and is **not** renegotiated server-side — the `maxoutcount` negotiation at `console.py:135` is on the client `Console` path only — so the PR's override is genuinely effective, not a no-op.

### 3. End-to-end `ipmitool` SOL loopback
With a pty as `ipmi.sol.device` and a real `ipmitool -I lanplus … sol activate` client on the box:
- Console stream flows smoothly through `_sol_pump` (~110 contiguous lines, no corruption/reordering).
- Keystrokes typed into `ipmitool` round-trip back through the pump (input path intact).
- **`max_packet` tolerance characterized** — forcing a single 4 KB contiguous burst and instrumenting pyghmi to record the *actual on-wire SOL payload sizes*:

| max_packet | max on-wire payload | 4 KB burst delivered to ipmitool |
|---:|---:|---|
| 256 (default) | 256 | ✅ full |
| 384 | 384 | ✅ full |
| 512 | 512 | ✅ full |
| 768 | 768 | ✅ full |
| 1024 | 1024 | ❌ **0 bytes — session wedges** |

pyghmi honors the cap all the way to 1024, and **ipmitool 1.8.19 accepts oversized SOL packets up to 768 B (3× the default → ~⅓ the ACK round-trips) but breaks at 1024 B** — the oversized packet is never ACKed and the stop-and-wait sender wedges on it. So the knob is real and useful, and this is concrete support for keeping the default at 256 while letting operators who control their client raise it and test.

### Not covered by this run
Real over-the-WAN RTT/jitter (I tested localhost loopback, where the round-trip is sub-millisecond, so the throughput *ceiling* isn't stressed — but the read-starvation fix itself is proven at unit level in §1), a physical USB-serial adapter, and SOL clients other than ipmitool.

Net: the read-starvation / burstiness fix works and the `max_packet` lever behaves exactly as designed on real hardware. Happy to run anything else while I have the device.

<sub>Testing performed with Claude Code on a PiKVM V4; device restored to its original state afterward.</sub>

PR #231, opened 2026-07-06
hid: humanize the mouse jiggler (bezier moves + randomized interval)

mawadSur:
This implements the **curve-based, human-like jiggler** that @mdevaev suggested in #222 — offered as a ready-to-merge alternative to the minimal interval-randomizer there. Full credit to @spili-desu for the original idea and the `±25%` interval-randomization approach (kept here), and to @mdevaev for the Bézier direction (`ghost_sim.py` + the `j.py` notes).

Since @spili-desu mentioned they need #222 soon and would rather do curves in a follow-up, this is that follow-up in one shot — happy to defer, fold into #222, or reshape it however you prefer.

## What it does

Replaces the fixed-period zig-zag jiggle with a single **universal** gesture engine and jitters the wait interval, so the anti-idle activity is harder to fingerprint. **No new config keys** — everything derives from the existing `jiggler.interval`.

- **Randomized interval** — jittered by ±25% of `jiggler.interval`, re-rolled after every jiggle (`60 → [45,75]`, floored at 1 s).
- **Random gesture per jiggle:**
  - `move` — a cubic-Bézier glide to a nearby random point **and back**, with ease-in/out timing and an independent random arc on each leg;
  - `jitter` — a few micro-twitches that settle back;
  - `scroll` — a net-zero wheel nudge.
- Works in **both absolute and relative** mouse modes, through the existing `send_mouse_move_event` / `send_mouse_relative_event` / `send_mouse_wheel_event` entry points.
- **Net-zero displacement** on every gesture, so the operator's cursor never drifts.

One file, `+116/−11`.

## Validation

- Byte-compiles on Python 3.14.
- 20k-trial geometry check: max relative per-step delta **22** (well under the ±127 HID limit — no truncation), net drift **exactly 0**.
- Functional test driving the real methods: absolute moves return to base & stay within `MouseRange`; relative deltas ≤127 with zero net; scroll/jitter net-zero; dispatcher clean in both modes.
- **Tested on real hardware** — PiKVM V4 (CM4), kvmd 4.171, Python 3.14.5: the patched `BaseHid` imports inside the real kvmd package and drives correct HID event streams against the genuine `MouseRange`/`MouseDelta` + `evdev`. All checks green.

## Note on CI

The `testenv` build currently fails repo-wide on `error: target not found: libjpeg` — unrelated to this change. Upstream Arch's `libjpeg-turbo` dropped the bare `libjpeg` provide; the fix is `libjpeg → libjpeg-turbo` in `testenv/Dockerfile` (also flagged in #222). This PR's CI should pass once that lands.

<sub>Prototyped and validated with Claude Code.</sub>

### DIFF
--- a/kvmd/plugins/hid/__init__.py
+++ b/kvmd/plugins/hid/__init__.py
@@ -21,6 +21,8 @@
 
 
 import asyncio
+import math
+import random
 import time
 
 from typing import Final
@@ -30,6 +32,8 @@ from typing import AsyncGenerator
 from typing import Any
 
 from evdev import ecodes
+
+from ...logging import get_logger
 
 from ...yamlconf import Section
 from ...yamlconf import Option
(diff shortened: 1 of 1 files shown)


mdevaev:
The author of the initial PR did a lot of work and expressed a desire to implement it on his own. I would prefer not to replace that PR with an AI code, especially considering how confusing everything looks.

PR #237, opened 2026-08-03
atx: add optional active-low relay support

cschanzle:
Summary
This PR adds optional power_switch_inverted and reset_switch_inverted configuration parameters to the ATX GPIO plugin.

Motivation
Some relay boards commonly used with DIY PiKVM are active-low. The current implementation assumes active-high outputs, which causes active-low relay modules to remain energized at idle and release during a button press.

This patch makes the output polarity configurable while preserving the existing default behavior.

Compatibility
Existing configurations are unchanged.  Both new options default to False.  Only users with active-low relay hardware need to set the new options.

Tested on a Raspberry Pi 4 using an active-low relay HAT ([XYGStudy Raspberry Pi Expansion Board](https://www.amazon.com/dp/B01G05KLIE)) controlling ATX power and reset.  Override.yaml:

```
kvmd:
    atx:
        type: gpio
        power_switch_pin: 26
        power_switch_inverted: true
        reset_switch_pin: 20
        reset_switch_inverted: true
        # fyi, relay defaults are:  click_delay: 0.1, long_click_delay: 5.5
        click_delay: 0.5  # Asus Z87-PRO motherboard (or my relays) needs more time
```

### DIFF
--- a/kvmd/plugins/atx/gpio.py
+++ b/kvmd/plugins/atx/gpio.py
@@ -58,6 +58,9 @@ class Plugin(BaseAtx):  # pylint: disable=too-many-instance-attributes
         self.__power_switch_pin: Final[int] = c.power_switch_pin
         self.__reset_switch_pin: Final[int] = c.reset_switch_pin
 
+        self.__power_switch_inverted = c.power_switch_inverted
+        self.__reset_switch_inverted = c.reset_switch_inverted
+
         self.__click_delay:      Final[float] = c.click_delay
         self.__long_click_delay: Final[float] = c.long_click_delay
 
@@ -94,6 +97,9 @@ class Plugin(BaseAtx):  # pylint: disable=too-many-instance-attributes
             "reset_switch_pin": Option(27,  type=valid_gpio_pin),
             "click_delay":      Option(0.1, type=valid_float_f01),
             "long_click_delay": Option(5.5, type=valid_float_f01),
+
+            "power_switch_inverted": Option(False, type=valid_bool),
+            "reset_switch_inverted": Option(False, type=valid_bool),
         }
 
     async def sysprep(self) -> None:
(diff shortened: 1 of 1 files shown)


mdevaev:
Hello. I'm strongly against supporting such a relay due to the fact that it is falsely triggered at the Raspberry reboot stage. Finding the right relay is not difficult, and there will be disproportionately fewer problems for support and the user.

Do you have a good reason we you need to support hardware unsuitable for PiKVM purposes?

cschanzle:
Thanks for the quick feedback.

The reason I proposed this patch is that my relay board does not appear to have the failure mode you described.  I tested cold boot and reboot, and the relays never energize during boot.  The only issue I observed was after kvmd requested the GPIO lines, where the ATX plugin assumes active-high outputs, and energizes the relays.

This patch was intended to configure the output polarity correctly for those boards while preserving the existing default behavior.

Is there a specific stage during Raspberry Pi startup that you expect would still produce a false trigger, or is your concern based on a class of active-low relay boards in general?

mdevaev:
This is not a Raspberry problem, it concerns some relays. Even if your specific relays are working, this does not mean that all other relays will work here. Like, for me right now it's a choice between getting you to buy the right relays, or opening the way for a lot of problematic assemblies that I'll have to deal with later. So I need a very good reason to support it.

cschanzle:
Thanks for the clarification.  I understand your concern is the broader support burden for unsupported relay hardware rather than my specific relay board.

I purchased my relay board over seven years ago for a Raspberry Pi thermostat project.  That project was successful but was eventually retired, so when I decided to build a DIY PiKVM, it seemed reasonable to reuse what I already had on-hand.  This has been a fun learning experience on many levels, and I can't thank you enough for creating and maintaining such an amazingly capable open-source software and hardware solution.

I'll close the PR and carry my patch locally for the time being.  For the benefit of anyone who finds this discussion in the future and decide to maintain a similar local modification, I've put safeguards in place to automatically re-apply the patch after kvmd upgrades and kvmd is prevented from starting if the patch no longer applies cleanly.

mdevaev:
Thank you for understanding, your kind words and I'm sorry to disappoint you. I hope that with the supported relay you will find PiKVM useful.

PR #240, opened 2026-08-19
Two-way audio in the direct H.264 mode

### LINKED ISSUE #1487: support microphone in direct mode
**Is your feature request related to a problem? Please describe.**
support microphone in direct mode

**Describe the solution you'd like**

**Describe alternatives you've considered**
https://github.com/AlexxIT/go2rtc

**Additional context**

phedoreanu:
**On hold** pending the audio server discussion below — @mdevaev is right that the direct mode
shouldn't take the ALSA devices for itself, and I'd rather agree on the architecture before
rewriting anything.

Stacked on #241, which carries the level meter and the raw capture switch. Those work in the
`WebRTC` mode on their own and don't depend on any of this, so they're reviewable separately;
this PR shows their commits until #241 is merged.

What's left here is the direct-mode audio itself: KVMD-Media captures the HDMI audio from the
TC358743, resamples it when the host isn't at 48 kHz, encodes Opus and sends it over the
websocket that already carries the video; the microphone goes the other way and is played to
the USB gadget. libasound, libopus and libspeexdsp are bound with ctypes.

It works and it's tested on a V4 Mini, but as it stands both devices are opened directly, so
the direct mode and WebRTC can't have audio at the same time — with `kvmd-janus-static` running,
janus holds both devices permanently and the direct mode gets nothing at all. That's the part
that needs the intermediate audio server, and the capture/playback layer here is expected to be
replaced by a memsink client once its shape is agreed.

Closes pikvm/pikvm#1487 when it lands.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -97,6 +97,9 @@ depends=(
 	dnsmasq
 	ipmitool
 	"janus-gateway-pikvm>=1.3.0"
+	alsa-lib
+	opus
+	speexdsp
 	certbot
 	"raspberrypi-io-access>=0.9"
 	raspberrypi-utils
--- a/configs/os/sysusers.conf
+++ b/configs/os/sysusers.conf
@@ -34,6 +34,7 @@ m kvmd kvmd-pst
 m kvmd kvmd-nbd
 
 m kvmd-media kvmd
+m kvmd-media audio
 
 m kvmd-pst kvmd
 
--- a/kvmd/apps/_scheme.py
+++ b/kvmd/apps/_scheme.py
@@ -484,6 +484,16 @@ def make_config_scheme() -> dict:
                     "drop_same_frames": Option(0.0, type=valid_float_f0),
                 },
             },
+
+            "audio": {
+                "device":   Option("plughw:tc358743,0", type=valid_stripped_string, if_empty=""),
+                "tc358743": Option("/dev/kvmd-video",   type=valid_abs_path, if_empty=""),
+            },
+
+            "mic": {
+                "device":  Option("plughw:UAC2Gadget,0", type=valid_stripped_string, if_empty=""),
+                "latency": Option(0.1, type=valid_float_f01),
+            },
         },
 
         "pst": {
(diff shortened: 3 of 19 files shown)


mdevaev:
Sorry, but it shouldn't be like this. We need to create some intermediate audio server to make audio available at the same time with webrtc. For things like this, I’d prefer to discuss the architecture in advance. And I don’t really trust AI code in terms of reliability.

phedoreanu:
You're right, and testing on a V4 Mini backs it up. With `kvmd-janus-static` running, janus holds
both `hw:tc358743,0` and the UAC2 gadget permanently, so the direct mode gets nothing at all —
worse than the "waits for the device" caveat I wrote in the PR.

I checked whether ALSA could do the sharing without new code: `dsnoop` and `dmix` both work here
(two readers at ~48000 frames/s each, two concurrent writers). But the slave rate is pinned in the
config and a consumer asking for another rate is simply refused — `Rate doesn't match (requested
48000Hz, get 44100Hz)`, no resampling. Since the HDMI rate is chosen by the host and changes at
runtime, a pinned dsnoop would pitch-shift on a switch, and re-pinning needs every consumer to
close first. So config-only isn't a real answer and an intermediate server is.

Before I write anything else — does this shape match what you have in mind?

* uStreamer keeps ownership of both devices (it already has them) and publishes the captured audio
  to a memsink, the same way it publishes H.264; the janus plugin and KVMD-Media both become readers.
* The microphone goes the other way: consumers write frames to a second sink and uStreamer plays
  them to the gadget.
* One open question: publish raw PCM and let each consumer encode, or encode Opus once centrally?
  Once is cheaper (~0.5 ms per frame on a CM4) but fixes the codec parameters for everyone.

Happy to write this up as a design first and leave the implementation to you if you'd prefer.
I'm converting this PR and the docs one to drafts in the meantime.

mdevaev:
It is necessary to create a separate server (in C) that would handle audio devices and then send/receive the stream from Janus or kvmd-media. I’m not sure exactly how this should be implemented, and developing this feature right now will prevent me from releasing the camera, because the architecture of the janus plugin will have to be completely changed. Don’t get me wrong, but I have high requirements for this code (both in terms of design and implementation), because PiKVM is the last means of providing access when everything else doesn’t work. This must be designed very carefully and written strictly by hand, with a full understanding of what each line does.

PR #241, opened 2026-08-20
Web UI: the raw microphone capture mode

phedoreanu:
Split as asked. This one is only the raw mic switch now: the `setMicDevice()` bugfix went to
#242, the level meter to #243.

The browser applies noise suppression and automatic gain control to the microphone by default.
That is right for a headset, but it wrecks a line-in or anything musical: the gain pumps and the
suppressor eats whatever does not look like speech. The switch turns both of them off.

The echo cancellation stays on in both modes. The host audio is played by the same page, so
without it the host would hear its own sound back through the microphone.

The constraints are negotiated with the offer, so flipping the switch while the microphone is
running restarts the session. The other two streamers get a no-op stub and keep the row hidden.

### DIFF
--- a/web/kvm/index.html
+++ b/web/kvm/index.html
@@ -424,6 +424,16 @@
                   <select class="small" id="stream-mic-selector" style="width: 90%;" disabled></select>
                 </td>
               </tr>
+              <tr class="feature-disabled" id="stream-mic-raw">
+                <td>Raw microphone:</td>
+                <td>
+                      <div class="switch-box">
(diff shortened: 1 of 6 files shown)


mdevaev:
I like this one, but can you split it into a separate PR to keep granularity of changes? Let's start only from raw mic feature without setMicEnabled bugfix and vu meter.

phedoreanu:
Done — this is now only the raw mic switch. The `setMicDevice()` fix is #242 (using your version of it), and the level meter is #243.

PR #242, opened 2026-08-21
Web UI: don't lose the microphone when the video mode is switched

phedoreanu:
Switching the video mode (or just loading the page with `WebRTC` saved as the mode) builds a
fresh `JanusStreamer` and then applies the saved UI state to it. At that moment the features
message has not arrived yet, so `__has_mic` is still `false` and `setMicDevice()` dropped the
request without a trace: the switch in the menu stays on, but the microphone is not sent until
you toggle it off and on by hand.

The fix remembers the last request and replays it once the features are known. The `WATCH` is
not sent at that point yet, so the device is applied in place and the session that has just
started is not restarted; only a device that disappeared in the meantime still needs a new
offer, which is the path that already existed.

The camera has the same shape of bug, but I left it alone: it needs a bit more thought because
`__camera_req` is filled from the features message itself.

### DIFF
--- a/web/share/js/kvm/stream_janus.js
+++ b/web/share/js/kvm/stream_janus.js
@@ -138,18 +138,23 @@ export function JanusStreamer(__setActive, __setInactive, __setInfo, __watchHook
 	};
 
 	self.setMicDevice = function(mic, reload=false) {
-		if (__has_mic && (__use_mic !== mic)) {
+		// The choice is remembered even before the features are known: this happens
+		// right after switching the video mode, when the UI applies its state to the
+		// fresh streamer. Only the restart needs the working session
+		if (__use_mic !== mic) {
 			if (mic) {
 				__refillDevices("mic", mic, function(id) {
 					__use_mic = id;
-					__destroyJanus();
+					if (__has_mic) {
+						__destroyJanus();
+					}
 				});
 			} else {
 				__use_mic = null;
 			}
 			reload = true;
 		}
-		if (reload) {
+		if (reload && __has_mic) {
 			__destroyJanus();
 		}
 	};


mdevaev:
Maybe it would be better option instead all of this? Could you test?
```js
diff --git a/web/share/js/kvm/stream_janus.js b/web/share/js/kvm/stream_janus.js
index 3ad69f8c..d2fe8a4c 100644
--- a/web/share/js/kvm/stream_janus.js
+++ b/web/share/js/kvm/stream_janus.js
@@ -138,18 +138,20 @@ export function JanusStreamer(__setActive, __setInactive, __setInfo, __watchHook
 	};
 
 	self.setMicDevice = function(mic, reload=false) {
-		if (__has_mic && (__use_mic !== mic)) {
+		if (__use_mic !== mic) {
 			if (mic) {
 				__refillDevices("mic", mic, function(id) {
 					__use_mic = id;
-					__destroyJanus();
+					if (__has_mic) {
+						__destroyJanus();
+					}
 				});
 			} else {
 				__use_mic = null;
 			}
 			reload = true;
 		}
-		if (reload) {
+		if (reload && __has_mic) {
 			__destroyJanus();
 		}
 	};
```

phedoreanu:
Yours is better, thanks — adopted it, the PR now carries your diff.

Tested it against a fake Janus plugin driving the real module, so the whole path from
`__resetStream()` to the first `WATCH` runs. Mimics what the UI does on a mode switch:
fresh streamer, saved state applied, then the features message arrives.

| variant | Janus sessions | mic in the first WATCH |
|-----------------|----------------|------------------------|
| current master  | 1              | no                     |
| your version    | 1              | yes                    |

The edge cases behave too: the mic switched off before the features stays off, and a uStreamer
that reports no mic still sends `mic: false` even though `__use_mic` is now set — `__sendWatch()`
and `getName()` both gate on `__has_mic`, so nothing leaks through.

One thing worth knowing: if a *specific* device is selected rather than the default one, it takes
two sessions instead of one (`WATCH` goes `mic: false` then `mic: true`). `__refillDevices()`
applies the default synchronously but has to wait for `listDevices()` for a named device, and by
then the features have usually arrived, so `__has_mic` is already true and the callback restarts
the session. It ends up correct either way, and it is the same restart that a device change
causes today, so I did not touch it.

I added a short comment above the check, since the reason the `__has_mic` test moved is not
obvious from the code. Drop it if you would rather keep the diff minimal.

PR #243, opened 2026-08-21
Web UI: the microphone level meter

phedoreanu:
Split out of #241 as asked.

There is no way to tell whether the host is receiving anything from the microphone without
checking the sound settings on the host itself. A muted OS input, a device with the gain at zero
and a working microphone all look the same in the menu.

The meter shows the level of the track that is actually being sent, after all the browser's
processing. The microphone is captured by Janus, so the meter only listens to the local track
from `onlocaltrack` and never opens the device itself. It reuses the existing progress widget,
so there is no new CSS and no new variables.

Tested in Chrome against a synthetic track: -15 dB at amplitude 0.25, -37 dB at 0.02,
`Clipping!` at 1.0, and the bar clears when the track goes away.

It touches the same few lines of the features handler as #241, so whichever goes in second needs
a trivial rebase.

Screenshot 2026-08-20 at 15 44 54

### DIFF
--- a/web/kvm/index.html
+++ b/web/kvm/index.html
@@ -434,6 +434,12 @@
                 </td>
                 <td class="tip">No noise suppression and no auto gain</td>
               </tr>
+              <tr class="feature-disabled" id="stream-mic-level">
+                <td>Mic level:</td>
+                <td class="slider" colspan="2">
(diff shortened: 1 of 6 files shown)


mdevaev:
I tried to resolve conflict but failed for some reason. Please rebase it.

phedoreanu:
Rebased.

mdevaev:
Thanks. I'll merge it, but hold the release, it requires a massive refactoring.

PR #244, opened 2026-08-22
otg: configurable camera resolution and isochronous bandwidth (1080p opt-in)

phedoreanu:
Rebased onto 4.213. The one-liner is gone, this is now a proper opt-in for 1080p that keeps the stock behaviour untouched by default.

### What changed

- **`otg.devices.camera.resolution`** (default `1280x720`) picks which frame the host is offered as its default, via `bDefaultFrameIndex`. All frames stay advertised, so a guest app that wants 640x480 still gets it; only the default moves. It works together with `safe`: when safe mode filters the requested resolution out, the kernel default is kept. A typo fails loudly at gadget setup instead of silently falling back.
- 1920x1080 is uncommented in the frame table. It only becomes the default when asked for.
- The frames now also get `dwDefaultFrameInterval`, `dwMinBitRate` and `dwMaxBitRate`. Without them the kernel keeps its own defaults, which describe a 640x360 frame at 15 FPS for every entry, no matter what was configured.
- **`otg.devices.camera.streaming.maxpacket` / `interval`** with the current 1024/4 as defaults. Those give about 1 MB/s of isochronous budget (maxpacket bytes per 2^(interval-1) microframes). I measured stock 4.213 on a V4 Mini with a plain webcam: 720p comes through at **15.7 fps** with 62 KB frames, 0.95 MB/s, right at that cap. So 720p30 is not reachable with the defaults either, and a detailed picture makes it worse; that is the "large frames don't fit into the time slot" effect from the comments above. A device that is set up per the docs (`isolcpus`, pinned UDC interrupt) can opt into interval 1:

  ```yaml
  otg:
      devices:
          camera:
              enabled: true
              resolution: 1920x1080
              streaming:
                  maxpacket: 1024
                  interval: 1
  ```
- Web: the sender tuning in `onlocaltrack` ran before the WebRTC negotiation finished, when `getParameters()` returns no encodings, so `setParameters()` was rejected and even the existing `degradationPreference` never applied. It now retries until the encodings exist and also raises the sender bitrate cap to 8 Mbps, because the browser default (2.5 Mbps on Chrome, about 1 Mbps on Firefox) starves 1080p30 and the camera looks like upscaled SD whatever the resolution says.
- Web: a webcam that does not have the exact mode the guest asked for used to fail with `OverconstrainedError`, and the handler only reset the device to the default one, which fails the same way, so the page kept creating and destroying a Janus session every second while the guest saw black. Now the constraints are relaxed to `ideal` on the first such error, and since the janus plugin drops frames of any other size than the one the guest expects, the captured track is scaled through the same canvas the mobile auto-rotate uses. A 720p webcam gives a soft 1080p picture instead of nothing.
- Web: the camera row in the System menu shows what is actually being sent, like `1920x1080@30`, taken from the `outbound-rtp` stats.

### On the missed requests

I looked at where they come from, since that is what made 1080p unstable. dwc2 on the CM4 runs in buffer DMA mode, so an isochronous IN endpoint gets exactly one request per (micro)frame and the next one is started from the completion interrupt. Whenever that interrupt is late, or the DMA has not filled the TX FIFO before the IN token arrives, the request comes back with `-ENODATA` and nothing was sent. Patch 1301 in linux-rpi-pikvm maps that onto `UVC_QUEUE_DROP_INCOMPLETE`, so a single missed 125 us window throws away the whole video frame. With bInterval 1 there are 8000 of those windows a second, which is why the misses "add up until the stream breaks", and why larger frames make it worse: more windows carry data, so more misses hit a payload instead of a harmless zero-length packet.

Measured on the V4 Mini (6.12.105-1-rpi, ucamera 0.5), 1080p at 2048/1 with the stock module: **7.5 fps at the gadget, 25 misses a second**, every one of them a lost frame.

For MJPEG the host does not care which microframe a payload arrives in, it just reassembles payloads until the FID bit toggles. So the payload can simply be sent again in the next window instead of being dropped. The kernel side is pikvm/packages#16, a lenient mode for isochronous IN in dwc2 (buffer DMA): a transfer that missed its frame is left armed or started late and goes out in the next possible (micro)frame, the way dwc2 behaved before the 2021 "Fix ISOC flow" commit, instead of being aborted with a 200 us poll in the interrupt handler and reported as `-ENODATA`. Module parameter `dwc2.isoc_in_lenient`, on by default. With it the same 1080p stream shows zero missed windows, whole frames, no corruption; the remaining fps limit is the webcam. This PR does not depend on it, the defaults stay 1024/4 either way; the two together are what makes 1080p work.

### Testing

V4 Mini, kvmd 4.213 with this branch, kernel 6.12.105-1-rpi, ucamera 0.5, macOS guest, Firefox and Chromium clients. The guest picks 1920x1080 by default, the resolution readout follows the actual stream, and a 720p-only webcam ends up as a scaled 1080p stream instead of a black picture. The numbers above are from `strace` on ucamera's `VIDIOC_DQBUF` and the `missed xfer` lines in dmesg.

### DIFF
--- a/kvmd/apps/_scheme.py
+++ b/kvmd/apps/_scheme.py
@@ -590,6 +590,10 @@ def make_config_scheme() -> dict:
                     "start":   Option(True,  type=valid_bool),
                     "safe":    Option(False, type=valid_bool),
                     "resolution": Option("1280x720", type=valid_stream_resolution),
+                    "streaming": {
+                        "maxpacket": Option(1024, type=valid_number.mk(min=1, max=3072)),
+                        "interval":  Option(4,    type=valid_number.mk(min=1, max=16)),
+                    },
                     "controls": {
                         "ct_mask": Option(0x00000E, type=valid_number.mk(min=0, max=0xFFFFFF)),
                         "pu_mask": Option(0x175B,   type=valid_number.mk(min=0, max=0xFFFF)),
--- a/kvmd/apps/_scheme.py
+++ b/kvmd/apps/_scheme.py
@@ -589,6 +589,7 @@ def make_config_scheme() -> dict:
                     "enabled": Option(False, type=valid_bool),
                     "start":   Option(True,  type=valid_bool),
                     "safe":    Option(False, type=valid_bool),
+                    "resolution": Option("1280x720", type=valid_stream_resolution),
                     "controls": {
                         "ct_mask": Option(0x00000E, type=valid_number.mk(min=0, max=0xFFFFFF)),
                         "pu_mask": Option(0x175B,   type=valid_number.mk(min=0, max=0xFFFF)),
--- a/kvmd/apps/otg/__init__.py
+++ b/kvmd/apps/otg/__init__.py
@@ -119,11 +119,12 @@ class _GadgetConfig:
 
         _mkdir(meta_path)
 
-    def add_camera(  # pylint: disable=too-many-locals
+    def add_camera(  # pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals
         self,
         starter: list[str],
         start: bool,
         safe: bool,
+        resolution: str,
         ct_mask: int,
         pu_mask: int,
     ) -> None:
@@ -158,9 +159,14 @@ class _GadgetConfig:
         func = "uvc.usb0"
         func_path = self.__create_function(func)
 
+        def interval(fps: int) -> int:
+            return math.floor(1 / fps * 10_000_000)  # 30 -> 333333, 100ns units
+
         _mkdir(join(func_path, "streaming/mjpeg/m"))
-        for (width, height, framerates) in [  # TODO: Make it configurable
-            # (1920, 1080, [30, 24, 20, 15, 10, 5]),
+        index = 0
+        default_index = 0
+        for (width, height, framerates) in [
+            (1920, 1080, [30, 24, 20, 15, 10, 5]),
             (1280, 720,  [30, 24, 20, 15, 10, 5]),
             (1024, 768,  [30, 24, 20, 15, 10, 5]),
             (1024, 576,  [30, 24, 20, 15, 10, 5]),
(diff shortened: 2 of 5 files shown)


mdevaev:
None of this is even close to our solution. I would ask you to wait for our release and then play around with it, because it doesn’t fit at all into the architecture we’ve already implemented. We’re going to do it at the beginning of the week.

phedoreanu:
I figured, from the direct audio PR. It is more for a PoC for 1080p@30fps, since the official (unreleased) webcam is going to be (for the beginning) only 720p.

mdevaev:
Right now: https://docs.pikvm.org/camera

mdevaev:
This is disabled because it doesn’t work stable in all cases according buy or tests. At some point, the number of missed requests will increase so much that the stream will be disrupted.

mdevaev:
I’ve just received a few bug reports stating that 720p sometimes also fails. It seems, the cause is the physical source camera. Some cameras produce a very detailed image, which, when transcoded, creates very large frames that don’t fit into the time slot of USB isochronous transfers. If you’d like to help figure out this problem (I’d really appreciate it), find another camera and try to investigate the issue in the kernel.

mdevaev:
Changing from 1024 to 2048 won’t work at all, because ucamera specifically wants 1024. All other changes are pointless if we can’t get 1080p to work correctly (I couldn’t).

In fact, the problem is that when decoding h264, it creates too much load on the memory bus, and this somehow affects the interrupt controller, which doesn’t have time to process the request at 2048/1. The reduction to 1024/4 was an attempt to reduce the load (successfully). We’re simply hitting the limits of the hardware, and no amount of maneuvering around the buffers will help here unless you know how to improve the interrupt controller’s performance.

I’ll be glad to take a look at your patch for UVC gadget to see if it works, but I’m somewhat skeptical (given the described issues with the interrupt controller).

phedoreanu:
@mdevaev I took you up on the kernel investigation, and I think I found where the missed requests come from and why big frames make it worse.

dwc2 on the CM4 is in buffer DMA mode, so an isochronous IN endpoint gets one request per microframe and the next one is started from the completion interrupt. When that interrupt is a bit late, or the DMA has not filled the TX FIFO before the IN token, dwc2 gives the request back with `-ENODATA` and nothing was sent. Your 1301 patch turns that into `UVC_QUEUE_DROP_INCOMPLETE`, so one missed 125 us window costs the whole video frame. The miss rate itself does not depend on the frame size, but the share of windows that carry data does, so a detailed picture turns more of the misses into lost frames. On my V4 Mini a 1080p stream at 2048/1 with the stock module came through at 7.5 fps with 25 misses a second.

For MJPEG the host does not care which microframe a payload lands in, it just reassembles until the FID bit flips, so the missed payload can simply be sent again in the next window. That is pikvm/packages#16: two requests at the UDC, on `-ENODATA` the one behind is pulled back with `usb_ep_dequeue()` and the missed one goes in front of it again. Same stream, same board: whole frames, and the remaining fps limit is the webcam. It is a module parameter, and it works best with `streaming_maxpacket=1024`, so a request is a single packet that either goes out whole or not at all.

This PR is reworked accordingly and no longer touches the defaults: 1024/4 and 720p stay as they are, 1080p is an opt-in through `resolution` and `streaming` in the override. It also fixes the browser side failing forever when the webcam does not have the exact mode the guest asked for. Numbers and details are in the description, also for the stock 4.213 config, where 720p tops out at about 16 fps on the 1 MB/s budget.

I only have this one V4 Mini and two webcams, so if you can try the kernel patch with the cameras that produced the bug reports, that would be the real test.

phedoreanu:
Update on the kernel side: the resend-from-UVC approach turned out to be wrong, so I replaced it. dwc2 discards a real XferCompl when a NAK from an earlier window is still pending, so the UVC function got `-ENODATA` for packets that had actually gone out and the resend duplicated data inside frames (corrupted JPEGs about once a second on a 1080p camera). On top of that every miss goes through `dwc2_hsotg_ep_stop_xfr()`, which polls up to 200 us in the interrupt handler and times out constantly here, making the next window late as well.

pikvm/packages#16 now fixes it in dwc2 itself: a lenient mode for isochronous IN that leaves a late transfer armed and lets it go out in the next (micro)frame, which is what the driver did before the 2021 "Fix ISOC flow for BDMA and Slave" commit. With that module and this branch at 1024/1, a 1080p webcam streams with zero missed windows, zero stop_xfr timeouts and whole frames on the guest. Description updated.

